PR #28, opened 2016-11-30
Filter methods without "operationId" from Martian instance.

markdingram:
Martian can now be used when not all methods have an operationId, allowing
a more gradual introduction of Martian.

### DIFF
--- a/core/src/martian/swagger.cljc
+++ b/core/src/martian/swagger.cljc
@@ -45,33 +45,35 @@
             %) parts)))
 
 (defn- ->handler [{:keys [definitions] :as swagger-map} url-pattern [method swagger-definition]]
-  (let [path-parts (tokenise-path url-pattern)
-        uri (string/join (map str path-parts))
-        parameters (:parameters swagger-definition)]
-    {:path uri
-     :path-parts path-parts
-     :method method
-     :path-schema (path-schema definitions parameters)
-     :query-schema (query-schema definitions parameters)
-     :body-schema (body-schema definitions parameters)
-     :form-schema (form-schema definitions parameters)
-     :headers-schema (headers-schema definitions parameters)
-     :response-schemas (response-schemas definitions (:responses swagger-definition))
-     :produces (some :produces [swagger-definition swagger-map])
-     :consumes (some :consumes [swagger-definition swagger-map])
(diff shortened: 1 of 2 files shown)


markdingram:
The Swagger spec is generated by compojure-api which doesn't add default values for operationId. This PR allows getting straight away to work with Martian, rather than blocking until every method is updated with an operationId.

oliyh:
Hi @markdingram,

Thanks for the PR! I am keen to make Martian usable in this scenario. However, as far as I am aware it should already work - you can't use the routes that don't have an operationId, but they don't break anything either.

Do you have a use case that doesn't work?

core/src/martian/swagger.cljc
```diff
@@ -45,33 +45,35 @@
...
-     :form-schema (form-schema definitions parameters)
-     :headers-schema (headers-schema definitions parameters)
-     :response-schemas (response-schemas definitions (:responses swagger-definition))
-     :produces (some :produces [swagger-definition swagger-map])
-     :consumes (some :consumes [swagger-definition swagger-map])
-     :swagger-definition swagger-definition
-     ;; todo path constraints - required?
-     ;; :path-constraints {:id "(\\d+)"},
-     ;; {:in "path", :name "id", :description "", :required true, :type "string", :format "uuid"
-     :route-name (->kebab-case-keyword (:operationId swagger-definition))}))
```
markdingram: my use case: (->kebab-case-keyword nil) throws, when first processing the swagger spec.

could replace with (some-> (:operationId swagger-definition) ->kebab-case-keyword), but I'm not sure how useful having an entry with no route-name is, hence my initial thought to filter out when no operationId.

markdingram:
Hi @oliyh, I added a comment to the code, the kebab casing method throws on a nil argument.

core/src/martian/swagger.cljc
oliyh: Thanks, I understand. I'll get this merged soon and released as a snapshot.

oliyh:
Hi @markdingram,

This feature is now present in martian 0.1.3-SNAPSHOT.

Thanks for your help!

markdingram:
Great! I'll try it.

Issue #5, opened 2016-06-14
Body params should be taken from supplied values and should not be named


oliyh:
Currently martian attempts to give the body params a name and expects values in a submap under that name, e.g. `(request-for m {:pet {:name "charlie" :age 2}})` where `:pet` is derived from the schema.

It should not do this; it is not robust and does not fit with the aim of abstracting HTTP structure from the function call. Instead, it should allow this: `(request-for m {:name "charlie" :age 2})`.

markdingram:
Hi @oliyh, either I'm missing something (which is entirely possible!), but wrapping the body params still seems to be required when loading from a http swagger JSON,

Referring back to:
https://juxt.pro/blog/posts/martian.html


```
(-> (martian-http/bootstrap-swagger "https://pedestal-api.herokuapp.com/swagger.json")
    (martian/request-for :create-pet {:name "Doggy McDogFace" :type "Dog" :age 3}))
```

Gives:
```
ExceptionInfo Value cannot be coerced to match schema: {:pet missing-required-key}  schema.coerce/eval7131/coercer!--7136/fn--7137/fn--7138 (coerce.clj:52)```
```

Wrapping with the schema pet works:
```
(martian/request-for :create-pet {:pet {:name "Doggy McDogFace" :type "Dog" :age 3}}))
```

(running in a REPL from current master a6c9cc75fbbb7510d722a01945a7e5f0a11486e0)

I can see the test case still passes, so is there a subtle difference between data driven & http derived martian?

Thanks!

oliyh:
Hi Mark,

That behaviour changed in order to accommodate vector bodies in #26, as it
was essentially destructuring the body param as a map which is not valid
for arrays.

The blog post is out of date, and actually the readme is too, thanks for
pointing it out. You should be using `:pet` as the parameter as you've
already worked out.

You can inspect individual routes by using the `explore` function or just
directly inspecting the Martian record if you think something has been
interpreted strangely.

markdingram:
Ahh thanks.

The underlying reason I was looking here was that in my compojure-api derived swagger there are some anonymous body schemas that end up being named some random variant of Body12345, Body45245 etc.

Like you mention it is no problem to pull out the name from Martian e.g.:

```
(-> (martian/explore m :create-pet) (:parameters) (first) (key))
```

...but feels a little clunky.

I wonder if a `:martian.core/body` key could be added that always resolves to the body schema? If you like the approach or suggest something better I can do the PR.

```
[...]

PR #30, opened 2016-12-08
Allow :martian.core/body to be used as the body schema name.

markdingram:
If no :martian.core/body parameter fallback to previous behaviour of
matching the schema name.

This change makes it easy to call methods with anonymous body schemas from
compojure-api which don't have deterministic names (body34512 etc).

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,11 @@ like that provided by [pedestal-api](https://github.com/oliyh/pedestal-api):
   (let [pet-id (-> (martian/response-for m :create-pet {:pet {:name "Doggy McDogFace" :type "Dog" :age 3}})
                    (get-in [:body :id]))]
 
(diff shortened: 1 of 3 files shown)


oliyh:
Thanks @markdingram I have pushed this to clojars with the latest snapshot.

ikitommi:
Hi, planned to throw away the generated top-level schemas names from Ring-Swagger, as swagger-ui now supports inline body schemas (was not the case 2y ago). I guess Martial can handle those via #5? So, instead of `Body34512` etc, would look like:

https://gist.githubusercontent.com/ikitommi/a5e9b6b7380ca470b66c591b85584110/raw/96abf1043ad040a2ac38d2d1b98352334f5aae0d/swagger.json

oliyh:
@ikitommi sounds good, I suppose there is an argument it also provides deduplication but in practise it's probably irrelevant. Martian can cope with those fine.

PR #53, opened 2018-06-18
Support trimming trailing slashes from base URL

samuelwagen:
Some Swagger definitions have "/" as their baseURL (since that is what
the Swagger specification demands) but that breaks Martian - if the base
URL is a single slash, and the individual operations URIs start with a
slash, then Martian emits double slashes in the final URL, i.e.
http://host//some-call.

This adds an optional trim-base-url? boolean to
cljs-http/bootstrap-swagger that, if enabled - removes a trailing slash
from the baseURL before passing it to core/bootstrap-swagger

I am not sure whether this is the right approach, but it seems to work.


oliyh:
Hi I'm having problems with the build on CI as you have seen, I want to get that sorted before merging this

PR #55, opened 2018-06-22
Support trimming trailing slashes from base URL

oliyh:
Some Swagger definitions have "/" as their baseURL (since that is what
the Swagger specification demands) but that breaks Martian - if the base
URL is a single slash, and the individual operations URIs start with a
slash, then Martian emits double slashes in the final URL, i.e.
http://host//some-call.

This adds an optional trim-base-url? boolean to
cljs-http/bootstrap-swagger that, if enabled - removes a trailing slash
from the baseURL before passing it to core/bootstrap-swagger

### DIFF
--- a/cljs-http/src/martian/cljs_http.cljs
+++ b/cljs-http/src/martian/cljs_http.cljs
@@ -3,7 +3,8 @@
             [cljs.core.async :refer [<!]]
             [martian.core :as martian]
             [martian.interceptors :as i]
-            [tripod.context :as tc])
+            [tripod.context :as tc]
+            [clojure.string :as str])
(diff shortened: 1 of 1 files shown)


oliyh:
This replaces #53 which I managed to somehow close by mistake

oliyh:
Hi @samuelwagen 

Thank you for the PR, I have merged and deployed to Clojars in `0.1.9-SNAPSHOT`. I may make some changes to the implementation including normalising across the other http implementations but I will try to ensure it does not break the functionality.

Cheers

samuelwagen:
Thank you, much appreciated.

PR #60, opened 2018-11-21
add cljdoc.edn config

martinklepsch:
Follow up to #59.

Currently a README from one of the talks is pulled in and cljdoc's slug generation isn't yet (https://github.com/cljdoc/cljdoc/issues/120) smart enough to discern the two READMEs. 

The added `docs/cljdoc.edn` file also includes the README of the Re-frame module and adds a CI step to verify that all files referenced in `docs/cljdoc.edn` exist.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,9 @@ jobs:
           paths:
             - /usr/local/bin/phantomjs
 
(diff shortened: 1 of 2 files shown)


oliyh:
Thank you @martinklepsch!

PR #64, opened 2019-04-30
Dont clobber records in params

andrewmcveigh:


oliyh:
Hi @andrewmcveigh thank you for this. I'm sorry I didn't reply on the issue you opened, I wanted to wait until I had time to consider it properly and give an informed reply, but I haven't yet found it!

I was wondering where in your interceptor chain you are doing your serialisation, is it after or before Martian's built ins? I am assuming after, hence why you want Martian to leave them alone, but if you did it before would that not be a workaround to this issue?

Thanks

andrewmcveigh:
Hi @oliyh. No probs, thanks for getting back to me.

I actually didn't mean to make this PR on your repo (at least not yet). I assumed I was making it against our fork.

Anyway...

I'm using the below interceptor chain (more-or-less), replacing `martian`'s default `encode-body` and `coerce-response` interceptors with an encoder map that knows about these content-types that we want to en/decode.

```clojure
(let [encoder (fn [content-type]
                {:encode (partial encode content-type)
                 :decode (partial decode content-type)
                 :as :stream})
      encoders (assoc (encoders/default-encoders)
                      "application/n-quads" (encoder "application/n-quads")
                      "application/n-triples" (encoder "application/n-triples"))
      interceptors (conj martian/default-interceptors
                         (interceptors/encode-body encoders)
                         (interceptors/coerce-response encoders)
                         martian-http/perform-request)]
  (martian-http/bootstrap-swagger "http://uri" {:interceptors interceptors}))
```

Sure, I could just encode to a binary stream before even calling `martian/response-for` (if that's what you mean?), but that doesn't seem like a very nice or flexible workaround.

I don't think putting these interceptors earlier in the stack would have any effect. The call to `keywordize-keys` on `params` happens before any of the interceptor stuff kicks in.

Issue #62, opened 2019-04-16
Consider not using `clojure.walk` to `keywordize-keys`


andrewmcveigh:
We're trying to use a custom de/serializer that knows how to de/serialize Records and lists of Records. However, if I pass these Records in the `params` arg of `response-for`, the call to `keywordize-keys` recursively transforms any Record into a map, which the de/serializer doesn't recognize.

Arguably this issue is in `clojure.walk/keywordize-keys`, but perhaps walking and transforming the full nested data structure is unnecessary?

https://github.com/oliyh/martian/blob/master/core/src/martian/core.cljc#L59
https://github.com/oliyh/martian/blob/master/core/src/martian/core.cljc#L66
https://github.com/oliyh/martian/blob/master/core/src/martian/core.cljc#L79

andrewmcveigh:
Hi. I'm moving my post over here because I'm still working on that branch.

Hi @oliyh. No probs, thanks for getting back to me.

I actually didn't mean to make this PR on your repo (at least not yet). I assumed I was making it against our fork.

Anyway...

I'm using the below interceptor chain (more-or-less), replacing martian's default encode-body and coerce-response interceptors with an encoder map that knows about these content-types that we want to en/decode.

```clojure
(let [encoder (fn [content-type]
                {:encode (partial encode content-type)
                 :decode (partial decode content-type)
                 :as :stream})
      encoders (assoc (encoders/default-encoders)
                      "application/n-quads" (encoder "application/n-quads")
                      "application/n-triples" (encoder "application/n-triples"))
      interceptors (conj martian/default-interceptors
                         (interceptors/encode-body encoders)
                         (interceptors/coerce-response encoders)
                         martian-http/perform-request)]
  (martian-http/bootstrap-swagger "http://uri" {:interceptors interceptors}))
```

Sure, I could just encode to a binary stream before even calling martian/response-for (if that's what you mean?), but that doesn't seem like a very nice or flexible workaround.

I don't think putting these interceptors earlier in the stack would have any effect. The call to keywordize-keys on params happens before any of the interceptor stuff kicks in.

oliyh:
Ah, I see what you mean, this happens before all the interceptors.
Perhaps it could be done in a new first interceptor instead, which would then make it easier for you to replace with your own without the need for forking or redeffing or anything nasty?

andrewmcveigh:
Yes, making it removable/replaceable would be preferable to the other options.

andrewmcveigh:
[...]

PR #65, opened 2019-05-07
Extract keywordize-keys on params to interceptors

andrewmcveigh:
clojure.walk/keywordize-keys recursively clobbers map-like data
structures. This commit pulls out the call to keywordize-keys on the
params map into the `default-interceptors`. It retains the default
behaviour, but allows the user to optionally customize the behaviour.

Fix: #62

### DIFF
--- a/core/src/martian/core.cljc
+++ b/core/src/martian/core.cljc
@@ -9,7 +9,8 @@
             [schema.core :as s]))
 
 (def default-interceptors
-  [interceptors/set-method
+  [interceptors/keywordize-params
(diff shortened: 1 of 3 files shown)


oliyh:
Hi @andrewmcveigh thank you for this, I think this is a much neater solution. Cheers

oliyh:
This is now on Clojars in 0.1.10-SNAPSHOT

andrewmcveigh:
Awesome, this is working for us now. Thanks!

PR #75, opened 2019-12-19
Support for path item object parameters

### LINKED ISSUE #74: [Bug] Martian is not respecting Path Item Object's parameters
In the OpenAPI-Specification 2.0 (same as Swagger 2.0) you can nest parameters like this:
```json
{
  "paths": {
    "/api/pets/{id}": {
      "get": {
        "method": "get",
        "parameters": [
          {
            "name": "watch",
            "type": "string",
            "in": "query"
          }
        ]
      },
      "parameters": [
        {
          "name": "id",
          "type": "string",
          "in": "path",
          "required": true
        }
      ]
    }
  }
}

```

But `martian` is only looking into `parameters` inside `get` (or any other http verb)

Reference:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#pathItemObject

RafaeLeal:
Fixes #74

### DIFF
--- a/core/src/martian/swagger.cljc
+++ b/core/src/martian/swagger.cljc
@@ -46,11 +46,15 @@
             (keyword param-name)
             %) parts)))
 
-(defn- ->handler [{:keys [definitions] :as swagger-map} url-pattern [method swagger-definition]]
+(defn- ->handler
+  [{:keys [definitions] :as swagger-map}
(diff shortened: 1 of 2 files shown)


oliyh:
This and #76 are now on Clojars as `0.1.12-SNAPSHOT`

PR #77, opened 2019-12-20
Fixes parameter-keys for open maps

RafaeLeal:
https://github.com/oliyh/martian/pull/76 introduced a bug on `enrich-handler`
More specifically on `parameter-keys`, since it is calling `s/explicit-schema-key` on `s/Any`
and giving this error
```
Bad explicit key: schema.core.AnythingSchema@3c6f0b88
```

Making the client unusable (for this swagger.json)


oliyh:
Thank you for this, it's now on clojars `0.1.12-SNAPSHOT`

PR #81, opened 2020-04-19
Introduced clj-http-lite

deas:
Hi,

This `clj-http-lite` based client is almost a complete copy of the `clj-http` client at least passes all tests and also works with [k8s-api](https://github.com/deas/k8s-api) (even with `watch`). Will be massaging pluggability in `k8s-api` a bit more and provide a PR over there soon. Still thinking about the `watch` implementation a bit, but I don't consider that related to `clj-http-lite` in `martian` scope.

### DIFF
--- a/clj-http-lite/src/martian/clj_http_lite.clj
+++ b/clj-http-lite/src/martian/clj_http_lite.clj
@@ -1,0 +1,31 @@
+(ns martian.clj-http-lite
+  (:require [clj-http.lite.client :as http]
+            [cheshire.core :as json]
+            [martian.core :as martian]
+            [martian.interceptors :as interceptors]))
+
+(defn- prepare-response-headers [headers]
+  (reduce (fn [m [k v]] (assoc m (keyword k) v)) {} headers))
+
+(def perform-request
+  {:name ::perform-request
+   :leave (fn [{:keys [request] :as ctx}]
+            (assoc ctx :response (-> (http/request request)
+                                     ;; convert keys to keywords
+                                     (update-in [:headers] prepare-response-headers))))})
+
+(def default-interceptors
+  (concat martian/default-interceptors [interceptors/default-encode-body interceptors/default-coerce-response perform-request]))
+
+(defn bootstrap [api-root concise-handlers & [opts]]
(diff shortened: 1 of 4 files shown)


clj-http-lite/src/martian/clj_http_lite.clj
```diff
@@ -0,0 +1,31 @@
...
+                                     (update-in [:headers] prepare-response-headers))))})
+
+(def default-interceptors
+  (concat martian/default-interceptors [interceptors/default-encode-body interceptors/default-coerce-response perform-request]))
+
+(defn bootstrap [api-root concise-handlers & [opts]]
+  (martian/bootstrap api-root concise-handlers (merge {:interceptors default-interceptors} opts)))
+
+(defn bootstrap-swagger [url & [{:keys [interceptors] :as params}]]
+  ;; :json not supported by clj-htt-lite
```
oliyh, line 24: Hi,
This comment confuses me, it says :json is not supported but the next line uses it anyway?

oliyh:
Hi,

Many thanks for this, it's exciting to have such contributions. It looks good but I left one comment which I hope you can clarify.

Thanks

clj-http-lite/src/martian/clj_http_lite.clj
deas: `clj-http-lite` does not support `:as :json` (yet) is what I am trying to say. With swagger, we have to have json in `swagger-definition` so we create the thing right there in `clj_http_lite.clj` threading `:body` through `(json/parse-string keyword)`. ;)

oliyh: Thanks for the clarification. I think it would be better in that case to remove the `:as :json` and the comment. It can be delegated to clj-http-lite in the future when it supports it.

deas: You are right about the confusion. Removed the misleading map and massaged the comment.

oliyh:
Thank you again for this! I will get a release together soon.

oliyh:
Available in `0.1.12-SNAPSHOT`

PR #82, opened 2020-04-19
Remove implicit httpkit dependency from martian test

davidjameshumphreys:
Add intercetptors `remove-stack` which is equivalent to httpkit `go-async`.


oliyh:
Looks good, thanks - much better than dragging in httpkit.

oliyh:
I think martian-httpkit can be removed from martian-test/project.clj now as well?

oliyh:
Available in `0.1.12-SNAPSHOT`

oliyh:
> I think martian-httpkit can be removed from martian-test/project.clj now as well?

Ignore this, I see it's just a dev dependency now, but it was implicit in the code.

PR #87, opened 2020-05-23
Implement support for parameter refs

kpmg-jp-jnorton:
Implement support for parameter refs

### DIFF
--- a/core/src/martian/schema.cljc
+++ b/core/src/martian/schema.cljc
@@ -79,27 +79,24 @@
 
 (defn schemas-for-parameters
   "Given a collection of swagger parameters returns a schema map"
-  [definitions parameters]
+  [ref-lookup parameters]
   (->> parameters
        (map (fn [{:keys [name required] :as param}]
               {(cond-> (keyword name)
                  (not required)
                  s/optional-key)
-               (make-schema definitions param)}))
+               (make-schema ref-lookup param)}))
        (into {})))
 
-(defn- resolve-ref [definitions ref]
-  (some->> ref
-           (re-find #"#/definitions/(.*)")
-           second
-           keyword
-           definitions))
+(defn- resolve-ref [ref-lookup ref]
+  (let [[_ category k] (re-find #"#/(definitions|parameters)/(.*)" ref)]
+    (get-in ref-lookup [(keyword category) (keyword k)])))
 
 (def URI
   #?(:clj java.net.URI
      :cljs goog.Uri))
 
-(defn- schema-type [definitions {:keys [type enum format $ref] :as param}]
+(defn- schema-type [ref-lookup {:keys [type enum format $ref] :as param}]
   (cond
     enum (apply s/enum enum)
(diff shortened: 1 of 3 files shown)


core/src/martian/schema.cljc
```diff
@@ -85,21 +85,26 @@
...
+               (make-schema definitions {} param)}))
        (into {})))
 
-(defn- resolve-ref [definitions ref]
-  (some->> ref
-           (re-find #"#/definitions/(.*)")
-           second
-           keyword
-           definitions))
+(defn- resolve-ref [definitions parameters ref]
```
oliyh, line 91: It feels like 'definitions' and 'parameters' are two categories of the same thing. I'd be tempted to merge them, either directly (risk of key collision?) or to provide it as a map that looks like `{:definitions {...} :parameters {...}}` and call it `ref-lookup`

core/src/martian/schema.cljc
```diff
@@ -85,21 +85,26 @@
...
+  (if-let [d (re-find #"#/definitions/(.*)" ref)]
```
oliyh, line 92: Then on this line you could change the regex to be
`(let [[_ category k] (re-find #"#/(definitions|parameters)/(.*)" ref)]
  (get-in ref-lookup [(keyword category) (keyword k)]))`

I'm not sure if 'category' is the best word to describe it.

oliyh:
Hi,

Thanks so much for this contribution. You are on the right lines, I just left some feedback that would help simplify it. Some unit tests would be great, and then I'd be happy to merge it.

Cheers
Oliy

core/src/martian/schema.cljc
kpmg-jp-jnorton: Please review the most recent comment.  Thank you.

core/src/martian/schema.cljc
kpmg-jp-jnorton: Please review the most recent comment.  Thank you.

core/src/martian/swagger.cljc
```diff
@@ -47,23 +47,23 @@
...
 
 (defn- ->handler
-  [{:keys [definitions] :as swagger-map}
+  [{:keys [definitions parameters] :as swagger-map}
    path-item-parameters
    url-pattern
    [method swagger-definition]]
   (when-let [route-name (some-> (:operationId swagger-definition) ->kebab-case-keyword)]
-    (let [path-parts (tokenise-path url-pattern)
+    (let [ref-lookup {:definitions definitions :parameters parameters} path-parts (tokenise-path url-pattern)
```
oliyh, line 55: This should be on its own line, and could be `(select-keys swagger-map [:definitions : parameters])`

oliyh:
Hi,

This looks really good now, I just left one last minor comment.

Thanks

core/src/martian/swagger.cljc
kpmg-jp-jnorton: Completed.  Please review latest version.

Issue #51, opened 2018-05-31
Support OpenAPI v3


troglotit:
Hey! Nice stuff!
Unfortunately, my project uses OpenAPI v3, and YAML, it would be dope if martian supported that.

oliyh:
Hi @troglotit 
Sorry it's taken me so long to reply. I agree it would be good to support the latest version, I will try to find time to support it.

In the meantime, PRs are welcome!

RafaeLeal:
Hi
I'd like to work on that, but I think it will require some refactoring..
I was thinking to try to decouple the handler-generation from the swagger/openapi, so I'd probably add an internal data structure that would represent routes. Does that make sense?

oliyh:
Hi @RafaeLeal 

There is an internal representation - you can use it if you are talking to a system that doesn't have a Swagger representation, examples in the tests and in the readme under "No Swagger, no problem":
```
(martian/bootstrap "https://api.org"
                   [{:route-name :load-pet
                     :path-parts ["/pets/" :id]
                     :method :get
                     :path-schema {:id s/Int}}

                    {:route-name :create-pet
                     :produces ["application/xml"]
                     :consumes ["application/xml"]
                     :path-parts ["/pets/"]
                     :method :post
                     :body-schema {:pet {:id   s/Int
                                         :name s/Str}}}])
```

Note that this is used with `bootstrap` as opposed to the Swagger json which is consumed with `bootstrap-swagger`, which first calls `swagger/swagger->handlers` to transform the Swagger json into the internal representation. https://github.com/oliyh/martian/blob/master/core/src/martian/core.cljc#L113

So really what I would expect is needed here would be to enhance the existing `martian.swagger` namespace to be able to deal with v3 as well, and return the same martian representation. I've been meaning to add a spec for the internal data structure to give better error messages but not got around to it.

oliyh:
I have added a spec for the internal handlers in #78 which should help with implementation.

PR #90, opened 2020-08-09
Add openapi->handlers functionality

czan:
I'm keen to use Martian for a project at work, but the specs that I have to interact with are OpenAPI3 rather than Swagger2. I tried using a converted from OpenAPI to Swagger, but they ended up yielding sub-par schemas (lots of optional keys and maybes that didn't match required fields in the spec).

In order to make Martian fit our project better, I've put together this `martian.openapi` namespace. I'm not claiming that it's a full implementation of the spec, and in particular the "content types" thing was a bit hastily thrown together, but this seems to be sufficient for what I need, and I figured it might be helpful for someone else, too.

I took inspiration from the `martian.swagger` namespace, but no code is shared with it.

### DIFF
--- a/core/src/martian/openapi.cljc
+++ b/core/src/martian/openapi.cljc
@@ -1,0 +1,116 @@
+(ns martian.openapi
+  (:require [camel-snake-kebab.core :refer [->kebab-case-keyword]]
+            [clojure.string :as string]
+            [schema.core :as s]))
+
+(defn- tokenise-path [url-pattern]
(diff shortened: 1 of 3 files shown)


czan:
Presumably if this ends up merged it would resolve #51.

oliyh:
Hi,

Thank you! This looks great, #51 has been outstanding for ages. I will take a close look in a bit.

Cheers

PR #96, opened 2020-08-27
Implement support for registering multiple martians with re-frame

benedekfazekas:
By default if no custom id for the martian is supplied it is
registered under `::martian.reframe/default-id` and the parameters for
a request using this martian does not need to change.

If a martian registered with a custom instance id `init` should be
called as `(martian.re-frame/init "https://some.url" {::martian.re-frame/instance-id :custom-id})`
and then the parameters for a request shold contain a
`::martian.re-frame/instance-id` key referencing the id used at init.

### DIFF
--- a/re-frame/test/martian/re_frame_test.cljs
+++ b/re-frame/test/martian/re_frame_test.cljs
@@ -74,9 +74,7 @@
    (re-frame/reg-event-db
     ::create-pet-success
     (fn [db]
-      (is (= :create-pet (-> @(re-frame/subscribe [::martian/pending-requests])
-                             first
-                             first)))
+      (is (= :create-pet (ffirst @(re-frame/subscribe [::martian/pending-requests]))))
       db))
 
    (martian/init "http://localhost:8888/swagger.json")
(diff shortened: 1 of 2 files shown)


re-frame/test/martian/re_frame_test.cljs
```diff
@@ -74,9 +74,7 @@
    (re-frame/reg-event-db
     ::create-pet-success
     (fn [db]
-      (is (= :create-pet (-> @(re-frame/subscribe [::martian/pending-requests])
-                             first
-                             first)))
+      (is (= :create-pet (ffirst @(re-frame/subscribe [::martian/pending-requests]))))
```
oliyh, line 77: 👍

re-frame/test/martian/re_frame_test.cljs
```diff
@@ -104,7 +102,7 @@
 
      (re-frame/reg-fx
       ::martian/request
-      (fn [[_ operation-id & args]]
+      (fn [[_ _ operation-id & args]]
```
oliyh, line 105: Will need to note that this is a breaking change to anyone who has overridden this effect (e.g. in their own tests)

oliyh:
This is great, thanks!

PR #120, opened 2021-08-04
throw an exception when the route name is not found

AndreaCrotti:


core/src/martian/core.cljc
```diff
@@ -46,7 +46,11 @@
     :else m))
 
 (defn find-handler [handlers route-name]
-  (first (filter #(= (keyword route-name) (:route-name %)) handlers)))
+  (let [handler (first (filter #(= (keyword route-name) (:route-name %)) handlers))]
```
oliyh, line 49: Slightly redundant `let`?

core/test/martian/core_test.cljc
```diff
@@ -311,9 +311,27 @@
            (martian/request-for m :create-blob {::martian/body body})
            (martian/request-for m :create-blob {:blob body})))))
 
+(deftest missing-route-test
+  (let [m (martian/bootstrap "https://camels.org"
+                             [{:route-name :create-camel
```
oliyh, line 316: Could slim down this test by just giving an empty array for the routes

core/test/martian/core_test.cljc
```diff
@@ -311,9 +311,27 @@
...
+                                 :query-schema {:camelVersion s/Int}
+                                 :body-schema {:Camel {:camelName s/Str
+                                                       :camelTrain {:leaderName s/Str
+                                                                    (s/optional-key :followerCamels) [{:followerName s/Str}]}
+                                                       :anyCamel s/Any}}
+                                 :headers-schema {(s/optional-key :camelToken) s/Str}
+                               :form-schema {:camelHumps (s/maybe s/Int)}}])]
+    (try
+      (martian/url-for m :missing-route {:camel-id 1})
+      (catch Exception e
```
oliyh, line 329:
```suggestion
      (catch Throwable e
```

`Throwable` is defined at the top in a clj/cljs friendly way, this is currently failing on CI on js

oliyh (commented):
Thanks for this. I agree this is a good way to go, although it is a breaking change to the API so will need to ensure this goes in the release notes as such.

core/src/martian/core.cljc
AndreaCrotti: Yes fixed

core/test/martian/core_test.cljc
AndreaCrotti: Done

PR #123, opened 2021-11-12
add way to pass extra options when fetching the swagger

AndreaCrotti:
This allows to pass extra options to the http/get that does the fetching of the swagger file.

# TODO / questions

- [x] This function is defined separately for each HTTP library used, should it be changed in all of them?
- [x] add a test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@
 **/.cljs_nashorn_repl/
 **/nashorn_code_cache/
 /**/out/
+.clj-kondo/.cache
+/.lsp/.cache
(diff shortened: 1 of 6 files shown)


oliyh:
Thanks, we should implement this in all http library integrations for uniformity
Don't worry too much about a test.

Cheers

AndreaCrotti:
Ok I've added it everywhere @oliyh

AndreaCrotti:
Only issue with this change is that  it's now positional, another option would be instead  of doing

`[url & [opts get-swagger-opts]]`

just put the `swagger-opts` inside the opts with something like

`[url & [{:keys [... swagger-opts] :as opts}]`

And then I would have to dissoc `swagger-opts` when passing it around though.
Not sure which one is the best option, I guess if we are not going to add anything else anyway the positional solution might be fine.

oliyh:
This is now in `0.1.19-SNAPSHOT`

Issue #125, opened 2021-12-23
Support for Open API Spec in YAML instead of JSON


bombaywalla:
I was attempting to use `martian` for the Amazon Ads API (specifically Sponsored Products) but that seems to have a [Open API Spec in YAML](https://d3a0d0y2hgofx6.cloudfront.net/openapi/en-us/sponsored-products/2-0/openapi.yaml)., which does not seem to be supported by `martian`.  I converted that YAML spec to EDN using `io.forward/yaml`, and then fed that to `(martian/bootstrap-openapi amznapi-base-uri oas-edn martian-http/default-opts))` but that gives me a NPE (during printing).
```
Execution error (NullPointerException) at martian.openapi/process-responses$iter$fn (openapi.cljc:98).
Cannot invoke "clojure.lang.Named.getName()" because "x" is null
```

I'm new to OpenAPI and `martian`.  Happy to dig deeper and work on a PR with some guidance.

oliyh:
Hello,

Thanks for reporting this. Could you attach the edn and yaml files please? 

Thanks

bombaywalla:
Had to change the extensions of the files to `.txt` to get github to allow the upload.

[sponsored-products-yaml.txt](https://github.com/oliyh/martian/files/7771605/sponsored-products-yaml.txt)
[sponsored-products-edn.txt](https://github.com/oliyh/martian/files/7771609/sponsored-products-edn.txt)

bombaywalla:
On taking a closer look, it seems that the EDN that is generated from the YAML may be faulty.
For example, the following YAML snippet
```
  /v2/sp/campaigns:
    post:
      tags:
        - Campaigns
      operationId: createCampaigns
      summary: Creates one or more campaigns.
      parameters:
        - $ref: '#/components/parameters/clientHeader'
        - $ref: '#/components/parameters/profileHeader'
      requestBody:
        description: An array of campaigns.
        content:
          application/json:
            schema:
              type: array
              minItems: 0
              maxItems: 100
              items:
                $ref: '#/components/schemas/CreateCampaign'
      responses:
        207:
          description: Success.
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/CampaignResponse'
        401:
          $ref: '#/components/responses/Unauthorized'
    put:
      tags:
        - Campaigns
      operationId: updateCampaigns
      summary: Updates one or more campaigns.
      parameters:
        - $ref: '#/components/parameters/clientHeader'
[...]

PR #132, opened 2022-01-19
Add OpenAPI YAML definition support.

bombaywalla:
For CLJ only.

Addresses #125

### DIFF
--- a/clj-http-lite/project.clj
+++ b/clj-http-lite/project.clj
@@ -6,6 +6,7 @@
   :plugins [[lein-modules "0.3.11"]]
   :dependencies [[com.github.oliyh/martian :version]
                  [org.martinklepsch/clj-http-lite "0.4.3"]
+                 [clj-commons/clj-yaml "0.7.107"]
                  [cheshire "5.10.1"]]
   :profiles {:provided {:dependencies [[org.clojure/clojure "1.10.3"]]}
              :dev {:source-paths ["../test-common"]
--- a/clj-http-lite/project.clj
+++ b/clj-http-lite/project.clj
@@ -5,9 +5,7 @@
             :url "http://opensource.org/licenses/MIT"}
   :plugins [[lein-modules "0.3.11"]]
   :dependencies [[com.github.oliyh/martian :version]
-                 [org.martinklepsch/clj-http-lite "0.4.3"]
-                 [clj-commons/clj-yaml "0.7.107"]
-                 [cheshire "5.10.1"]]
+                 [org.martinklepsch/clj-http-lite "0.4.3"]]
   :profiles {:provided {:dependencies [[org.clojure/clojure "1.10.3"]]}
              :dev {:source-paths ["../test-common"]
                    :exclusions [[org.clojure/tools.reader]]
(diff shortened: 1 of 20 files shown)


clj-http-lite/project.clj
```diff
@@ -6,6 +6,7 @@
   :plugins [[lein-modules "0.3.11"]]
   :dependencies [[com.github.oliyh/martian :version]
                  [org.martinklepsch/clj-http-lite "0.4.3"]
+                 [clj-commons/clj-yaml "0.7.107"]
```
oliyh, line 9: Why not add the clj-yaml dependency to core? Without it, core will not technically compile on its own, and that is where the clj-yaml dependency actually is.

core/src/martian/yaml.clj
```diff
@@ -0,0 +1,32 @@
+(ns martian.yaml
```
oliyh, line 1: I wouldn't mind a test for this ns, just at a very basic level (single route)

clj-http-lite/project.clj
bombaywalla: Sure. Will do.

I guess I really don't know CLJS and was not sure if I could / should have a CLJ file in something that was CLJC.

core/src/martian/yaml.clj
bombaywalla: Okay

bombaywalla:
This PR is ready for review.

oliyh:
Hi,

I was looking at the http-kit module which has a require for Cheshire, so it should still have Cheshire in the project clj rather than relying on the user also having a dependency on core (even though they realistically would)

bombaywalla:
Sorry if I'm being dense, but...

The `project.clj` for `martian.http-kit` has a dep for `martian.core`, which has a dep for `cheshire`.
So, if a user used `martian.http-kit` (but did not explicitly include `martian.core`) they'd still get `cheshire`, right?

oliyh:
Hi @bombaywalla 
Yes, it would still get cheshire on the classpath but transitively. If martian.http-kit only made calls to martian.core, and martian.core made all the calls to cheshire I'd be fine with that, but martian.http-kit explicitly requires cheshire and makes calls to it. To mirror this there should be an explicit dependency on cheshire in martian.http-kit.

This will avoid the scenario where some developer removes all usage of cheshire in martian.core, and removes it from the core project.clj thinking it's no longer used. This would unintentionally and unknowingly break compilation of martian.http-kit.

Many thanks

oliyh:
Same for clj-http-lite as well. Hato and clj-http don't need it as those libraries natively convert json to clojure.

Thanks

bombaywalla:
Okay.  Makes sense.  Added `cheshire` back to `httpkit` and `clj-http`.
Please review.

oliyh:
Hi,

Sorry I made a half mistake in my previous comment - clj-http-lite *does* need cheshire, clj-http doesn't.

Once that's done I think this is good to go, thank you.

bombaywalla:
Fixed.

oliyh:
Thank you!

oliyh:
I'll build a snapshot to clojars tomorrow

oliyh:
Hi, this is now in `0.1.21-SNAPSHOT`

bombaywalla:
Thanks.

PR #136, opened 2022-01-30
Update README.md

rgkirch:
https://github.com/oliyh/martian/blob/master/core/src/martian/interceptors.cljc#L151
`inject` matches on the name


oliyh:
Thank you for this, you're quite right

Issue #143, opened 2022-02-06
Automatically handle application/vnd.blah.foo+json custom content types


bombaywalla:
Many OpenAPI documents appear to have custom vendored Content-Types that are of the form `application/vnd.blah.foo+json`.  One such example can be found at [Amazon Ads API Manager Account](https://dtrnk0o2zy01c.cloudfront.net/openapi/en-us/dest/ManagerAccount_prod_3p.json). These are all `JSON` but with extra constraints.  See [RFC 6839](https://datatracker.ietf.org/doc/html/rfc6839#section-3.1) for the `+json` specification.

Martian allows [custom content types](https://github.com/oliyh/martian#custom-content-types), but it is tedious to add all the custom types by hand.  Is there a recommended way to handle this situation?

If not, I propose that martian provide an optional way to treat vendored custom content types ending in `+json` to be the same as `application/json` for the purposes of encoding and decoding.  It might be good (but not required, IMO) to provide for a mechanism to override the encoders / decoders for specific `+json` content types if needed.

I am happy to work on a PR if this is deemed useful.

I can envision several approaches:

1. Provide a function that will scan an existing martian instance and return a new one with the additional encoders/decoders. I don't know the martian code well enough, but it appears that no information is lost from the OpenAPI definition and so this should be doable without impacting the existing code. The downside is that there would be duplicate implementations for each custom content type.  
2. Similar to the one above, but the custom content type handlers are (optionally) added dynamically as needed - if there is an already existing compatible supported content type.  So, for example, if `vnd.foo+json` is searched for and not found, then if `application/json` is supported, martian will add `vnd.foo+json` to the list of supported types, with the same implementation as `application/json`.
3. Checking to see if a particular content type is currently supported assumes an exact match between the content type and one of the supported content types.  If the supported content type were to be a regex, then one single encoder/decoder could handle multiple content types.  This would be a change in the semantics of martian, but could be backwards compatible.  There might be precedence issues when more than one regex matches.

Thoughts?

oliyh:
Hi,

In this situation, which seems quite specific to the services being called, I'm inclined to think it's a problem for the user to solve rather than martian: if the encoding/decoding interceptor provided is not sufficient, you can just write your own interceptor and use it instead.

[...]

PR #144, opened 2022-02-08
Add support for the :plus-json option to bootstrap-openapi.

bombaywalla:
This adds encoders and decoders for all content types of the form
application/vnd.foo.bar+json that are found in the OpenAPI spec.

Should address #143

### DIFF
--- a/core/src/martian/core.cljc
+++ b/core/src/martian/core.cljc
@@ -5,7 +5,7 @@
             [martian.interceptors :as interceptors]
             [martian.parameter-aliases :as parameter-aliases :refer [parameter-aliases alias-schema]]
             [martian.swagger :refer [swagger->handlers]]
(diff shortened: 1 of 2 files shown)


bombaywalla:
Sorry.  Commit 25e9314 had an error where I accidentally overwrote your recent change about expanding refs in parameters.  Commit 10c128c reverts that accidental change.

oliyh:
Hi,

Given the complexity of the code and what I consider to be a slightly esoteric API that creates the use case, I think coping with these content types is best done by the user with their own serialisation interceptors replacing or in addition to the martian ones. Hopefully there is plenty of code in martian that can be reused to help with this.

I appreciate your efforts here and in the other work you've done, so thank you for that but apologies I don't think this is a good fit for martian.

Thanks

bombaywalla:
Okay.  Fair enough.  I can easily keep it outside of martian.

Issue #145, opened 2022-02-11
Martian could be more specific in the schemas


bombaywalla:
For the following OpenAPI spec
```yaml
openapi: 3.0.2
info:
  version: "2.0"
  title: Testing Martian
servers:
  - url: "https://example.com"
security:
  - bearer: []
paths:
  /v1/campaigns:
    get:
      operationId: listCampaigns
      summary: Gets an array of campaigns.
      parameters:
        - $ref: '#/components/parameters/startIndex'
      responses:
        200:
          description: Success.
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Campaign'
        401:
          $ref: '#/components/responses/Unauthorized'
        404:
          $ref: '#/components/responses/NotFound'
components:
  parameters:
    startIndex:
      name: startIndex
      in: query
      schema:
        type: number
        default: 0
  responses:
    NotFound:
      description: The requested resource was not found.
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/Error' 
    Unauthorized:
      description: Unauthorized.
      content:
        application/json:
          schema:
            $ref: '#/components/schemas/Error'
  schemas:
    Campaign:
      properties:
        name:
          description: The name of the campaign.
          type: string
    Error:
      properties:
        details:
          description: A human-readable description of the error.
          type: string
```
martian generates the following:
```clojure
user> (def m (martian-http/bootstrap-openapi "http://localhost:8090/testsp.yaml" martian-http/default-opts))
#'user/m
user> (martian/explore m :list-campaigns)
{:summary "Gets an array of campaigns.",
 :parameters {{:k :start-index} (default Any 0)},
 :returns {200 [Any], 401 nil, 404 nil}}
```
I would have expected something closer to:
```clojure
user> (martian/explore m :list-campaigns)
{:summary "Gets an array of campaigns.",
 :parameters {{:k :start-index} (default Num 0)},
 :returns {200 [{{:k :name} java.lang.String}], 401 nil, 404 nil}}
```
And, there seems to be something wrong with the way martian handles the schema for the error codes.

From my reading of [this issue](https://github.com/OAI/OpenAPI-Specification/issues/1657) the above spec is valid.

bombaywalla:
PRs  #146,  #147 and  #148 address this issue completely.  So, closing it.

PR #148, opened 2022-02-13
Handle schemas with no type.

bombaywalla:
This partially addresses #145.

If a schema has no "type" key, and ONLY has a "properties" key, then
it is a valid schema and we can reasonably assume that the type is "object".

See https://github.com/OAI/OpenAPI-Specification/issues/1657

Excerpt:
A particularly common form of this is a schema that omits type, but specifies properties.
Strictly speaking, this does not mean that the value must be an object.
It means that if the value is an object, and it includes any of those properties,
the property values must conform to the corresponding property subschemas.

In reality, this construct almost always means that the user intends type: object,
and I think it would be reasonable for a code generator to assume this,
maybe with a validation: strict|lax config option to control that behavior.

### DIFF
--- a/core/src/martian/openapi.cljc
+++ b/core/src/martian/openapi.cljc
@@ -55,22 +55,21 @@
      (wrap schema
            (condp = (if-let [typ (:type schema)]
                       typ
-                      (let [ks (keys schema)]
-                        ;; If a schema has no :type key, and the only key it contains is a :properties key,
-                        ;; then the :type can reasonably be inferred as "object".
-                        ;;
-                        ;; See https://github.com/OAI/OpenAPI-Specification/issues/1657
-                        ;;
-                        ;; Excerpt:
-                        ;; A particularly common form of this is a schema that omits type, but specifies properties.
-                        ;; Strictly speaking, this does not mean that the value must be an object.
-                        ;; It means that if the value is an object, and it includes any of those properties,
-                        ;; the property values must conform to the corresponding property subschemas.
-                        ;;
-                        ;; In reality, this construct almost always means that the user intends type: object,
-                        ;; and I think it would be reasonable for a code generator to assume this,
-                        ;; maybe with a validation: strict|lax config option to control that behavior.
-                        (when (and (= (count ks) 1) (= (first ks) :properties)) "object")))
+                      ;; If a schema has no :type key, and the only key it contains is a :properties key,
+                      ;; then the :type can reasonably be inferred as "object".
+                      ;;
+                      ;; See https://github.com/OAI/OpenAPI-Specification/issues/1657
+                      ;;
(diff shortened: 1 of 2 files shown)


oliyh:
Hello,

Thanks for your continued work on this. From reading the spec, it seems like it could be expressed as:

```clj
(s/either {(s/optional k) v ...} ;; if it's an object and it includes _any_ of these properties
          s/Any ;; does not mean the value has to be an object
          )
```

It's not pretty, but seems to closer match the intention of the spec. What do you think?

bombaywalla:
What you have is more accurate with respect to the spec. I'd guess that in the particular example I am dealing with, the person writing the spec just meant for it to be an object, and didn't understand the subtleties.

Would you like me to modify the PR to make it what you suggest?

oliyh:
I'm inclined to agree with you, it seems a bit rubbish that it would allow `"foo"` through, and actually I don't think schema would give the right behaviour anyway. Consider:

Properties of `foo` and `bar` as ints in the schema, and the user supplies `{:foo 123 :bar "not a number"}`. This would pass the schema I wrote above, because it would fail the map schema but then pass the Any schema. From the wording, because the keys are present, they should use the map schema only and therefore fail.

Really, the 'correct' implementation is this:
```clj
(s/conditional 
  #(some? % [:foo :bar]) 
  {(s/optional-key :foo) s/Int 
   (s/optional-key :bar) s/Int} 

  :else s/Any)
```

But this seems to be getting overly complicated now. I think, given that no one has raised an issue about this before, that we go with your implementation and we can always adjust it later if it causes problems for someone else.

core/src/martian/openapi.cljc
```diff
@@ -53,7 +53,24 @@
...
+                        ;; Excerpt:
+                        ;; A particularly common form of this is a schema that omits type, but specifies properties.
+                        ;; Strictly speaking, this does not mean that the value must be an object.
+                        ;; It means that if the value is an object, and it includes any of those properties,
+                        ;; the property values must conform to the corresponding property subschemas.
+                        ;;
+                        ;; In reality, this construct almost always means that the user intends type: object,
+                        ;; and I think it would be reasonable for a code generator to assume this,
+                        ;; maybe with a validation: strict|lax config option to control that behavior.
+                        (when (and (= (count ks) 1) (= (first ks) :properties)) "object")))
```
oliyh, line 73: This whole `let` block could probably be simplified to `(= #{:properties} (set (keys schema)))`

bombaywalla:
Made the code improvement you suggested.

oliyh:
Many thanks once again

bombaywalla:
You're welcome  Thanks for martian and for guiding the PRs to successful merge.

PR #155, opened 2022-11-05
Support date-time and int-or-string string formats

jcdlbs:
These changes assist with using the Kubernetes API which makes use of date-time fields and has the concept of a int-or-string format. 

Without the int-or-string type - Martian gives a validation error when trying to use the ':port 8080' in a kubernetes readiness check complaining that a String is required. :port 8080 is a valid value as-is :port "https" and they have different meanings to Kubernetes.

I'm not sure if this specific use case makes sense for upstream and the tests for these enhancements could probably be expanded but I figured I'd submit this pull request just in case.

### DIFF
--- a/core/src/martian/schema.cljc
+++ b/core/src/martian/schema.cljc
@@ -85,6 +85,7 @@
     (= "string" type) (case format
                         "uuid" (s/cond-pre s/Str s/Uuid)
                         "uri" (s/cond-pre s/Str URI)
+                        "date-time" (s/cond-pre s/Str s/Inst)
                         s/Str)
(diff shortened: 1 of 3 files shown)


oliyh:
Thanks! The int-or-string does sound like quite a specific case but it won't collide with anything else so I'm happy to have it as part of the core library.

Thank you for the contribution, I will publish a snapshot build next week.

oliyh:
Hi, this is in `0.1.22-SNAPSHOT` on clojars now

jcdlbs:
Excellent, thanks!!

PR #164, opened 2023-02-14
add support for the deprecated field on the operation

wkok:
When an operation is deprecated in the openapi spec this PR makes the `deprecated` field available when exploring the endpoint.

### DIFF
--- a/core/src/martian/core.cljc
+++ b/core/src/martian/core.cljc
@@ -107,16 +107,16 @@
 (defn explore
   ([martian] (mapv (juxt :route-name :summary) (:handlers (resolve-instance martian))))
   ([martian route-name]
-   (when-let [{:keys [parameter-aliases summary deprecated] :as handler} (find-handler (:handlers (resolve-instance martian)) route-name)]
-     {:summary summary
-      :deprecated deprecated
-      :parameters (reduce (fn [params parameter-key]
-                            (merge params (alias-schema (get parameter-aliases parameter-key) (get handler parameter-key))))
-                          {}
-                          parameter-schemas)
-      :returns (->> (:response-schemas handler)
-                    (map (juxt (comp :v :status) :body))
-                    (into {}))})))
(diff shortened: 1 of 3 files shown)


core/src/martian/core.cljc
```diff
@@ -107,8 +107,9 @@
 (defn explore
   ([martian] (mapv (juxt :route-name :summary) (:handlers (resolve-instance martian))))
   ([martian route-name]
-   (when-let [{:keys [parameter-aliases summary] :as handler} (find-handler (:handlers (resolve-instance martian)) route-name)]
+   (when-let [{:keys [parameter-aliases summary deprecated] :as handler} (find-handler (:handlers (resolve-instance martian)) route-name)]
```
oliyh, line 110: Do you think `:deprecated?` is more conventional, for a Boolean?

oliyh:
Hi,

Thanks for the contribution! There are some failing tests in `martian.core-test` that need updating, and I've left a couple of comments too.

Cheers

core/src/martian/core.cljc
```diff
@@ -107,8 +107,9 @@
...
      {:summary summary
+      :deprecated deprecated
```
oliyh, line 112: It would be cleaner to only merge the deprecated key when it is true, to avoid cluttering up the explore output with a load of `:deprecated? false` (or, in the case of swagger, `nil`)

wkok:
Updated with your suggestions.
The tests that still fail seems to do with a host that's maybe not running?

wkok:
Updated with your suggestions.

The tests that still fail seems to be related to some host?
`ERROR - GET https://pedestal-api.herokuapp.com/swagger.json's callback `

oliyh:
That's an integration test, unfortunately I lost that app in the great heroku shutdown. It's ok to be red until I can get it running again somewhere else.

Thanks!

PR #165, opened 2023-03-13
Babashka support

borkdude:
This PR adds both babashka support for the core module and support for babashka.http-client.

### DIFF
--- a/babashka-http-client/test/martian/babashka/http_client_test.clj
+++ b/babashka-http-client/test/martian/babashka/http_client_test.clj
@@ -1,0 +1,124 @@
+(ns martian.babashka.http-client-test
+  (:require [martian.babashka.http-client :as martian-http]
+            [martian.core :as martian]
+            [martian.encoders :as encoders]
+            [martian.server-stub :refer [with-server swagger-url openapi-url openapi-test-url openapi-yaml-url openapi-test-yaml-url]]
+            [martian.test-utils :refer [input-stream->byte-array]]
+            [clojure.test :refer [deftest testing is use-fixtures]]))
+
+(use-fixtures :once with-server)
+
+(deftest swagger-http-test
+  (let [m (martian-http/bootstrap-swagger swagger-url)]
+
+    (testing "default encoders"
+      (is (= {:method :post
+              :url "http://localhost:8888/pets/"
+              :body {:name "Doggy McDogFace", :type "Dog", :age 3}
+              :headers {"Accept" "application/transit+msgpack"
+                        "Content-Type" "application/transit+msgpack"}
+              :as :byte-array
+              :version :http-1.1}
+             (-> (martian/request-for m :create-pet {:pet {:name "Doggy McDogFace"
+                                                           :type "Dog"
+                                                           :age 3}})
+                 (update :body #(encoders/transit-decode (input-stream->byte-array %) :msgpack))))))
+
+
+    (let [response (martian/response-for m :create-pet {:pet {:name "Doggy McDogFace"
+                                                              :type "Dog"
+                                                              :age 3}})]
+      (is (= {:status 201
+              :body {:id 123}}
+             (select-keys response [:status :body]))))
+
+    (let [response (martian/response-for m :get-pet {:id 123})]
+      (is (= {:name "Doggy McDogFace"
+              :type "Dog"
+              :age 3}
+             (:body response))))))
+
+(deftest async-test
+  (let [m (martian-http/bootstrap-swagger swagger-url {:interceptors martian-http/default-interceptors-async})]
+
+    (testing "default encoders"
+      (is (= {:method :post
+              :url "http://localhost:8888/pets/"
+              :body {:name "Doggy McDogFace", :type "Dog", :age 3}
+              :headers {"Accept" "application/transit+msgpack"
+                        "Content-Type" "application/transit+msgpack"}
+              :as :byte-array
+              :version :http-1.1}
+             (-> (martian/request-for m :create-pet {:pet {:name "Doggy McDogFace"
+                                                           :type "Dog"
+                                                           :age 3}})
+                 (update :body #(encoders/transit-decode (input-stream->byte-array %) :msgpack))))))
+
+
+    (let [response @(martian/response-for m :create-pet {:pet {:name "Doggy McDogFace"
+                                                               :type "Dog"
+                                                               :age 3}})]
+      (is (= {:status 201
+              :body {:id 123}}
+             (select-keys response [:status :body]))))
+
+    (let [response @(martian/response-for m :get-pet {:id 123})]
+      (is (= {:name "Doggy McDogFace"
+              :type "Dog"
+              :age 3}
+             (:body response))))))
+
+(deftest error-handling-test
+  (testing "remote exceptions"
+
+    (testing "can be thrown"
+      (let [m (martian-http/bootstrap-swagger swagger-url {:interceptors martian-http/default-interceptors-async})]
+        (is (thrown? Exception @(martian/response-for m :get-pet {:id -1})))))
+
+    (testing "can be handled"
+      (let [turn-off-exception-throwing {:name ::turn-off-exception-throwing
+                                         :enter (fn [ctx] (assoc-in ctx [:request :throw] false))}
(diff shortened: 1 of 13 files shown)


borkdude:
@oliyh I'm sure you will have some feedback and request for changes, let me know.

The frankiesardo/linked dependency is no longer needed, but I didn't know if you were alright with me removing it.

core/project.clj
```diff
@@ -16,7 +16,8 @@
 
                  [com.cognitect/transit-clj "1.0.324"]
                  [com.cognitect/transit-cljs "0.8.269"]
-                 [frankiesardo/linked "1.3.0"]]
+                 [frankiesardo/linked "1.3.0"]
```
borkdude, line 19: This dep could be removed in favor of the one below since the one below works in both bb and JVM

core/test/martian/test_helpers.cljc
```diff
@@ -5,8 +5,8 @@
 
 #?(:clj
    (defmacro json-resource [resource-name]
-     (json/parse-string (slurp (io/resource resource-name)))))
+     (list 'quote (json/parse-string (slurp (io/resource resource-name))))))
 
 #?(:clj
    (defmacro yaml-resource [resource-name]
-     (yaml/parse-string (slurp (io/resource resource-name)))))
+     (list 'quote (yaml/parse-string (slurp (io/resource resource-name))))))
```
borkdude, line 12: I needed to quote this since the EDN representation of a flatland linked list is written like `#ordered ([:a b])` etc. Not quoting that, would evaluate the data structure as a function call.

core/project.clj
oliyh: Happy to remove frankiesardo/linked

core/src/martian/encoders.cljc
```diff
@@ -37,15 +38,15 @@
   ([key-fn]
    (merge
     #?(:clj
-       {"application/transit+msgpack" {:encode #(transit-encode % :msgpack)
+       {"application/transit+json"    {:encode #(transit-encode % :json)
```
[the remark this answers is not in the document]
oliyh: The order is important here (although it looks accidental because of not using an ordered map for each of these cases), as it specifies an order of preference of encoders. If bb doesn't support this one, can we just have an extra `:bb` dispatch which doesn't include it?

core/test/martian/test_helpers.cljc
oliyh: I'm not sure I understand, these should just be normal json and yaml data using clojure's native data structures?

oliyh:
Hi @borkdude thanks for the PR. I've replied to most of the comments you left yourself, and will have a more detailed look through the rest of the code shortly but it looks good after a quick skim.

babashka-http-client/test/martian/babashka/http_client_test.clj
```diff
@@ -0,0 +1,152 @@
...
+    (is (contains? (set (map first (martian/explore myaml)))
+                   :get-order-by-id))
+
+    (is (= "http://localhost:8888/openapi/v3"
+           (:api-root m)))
+
+    (is (contains? (set (map first (martian/explore m)))
+                   :get-order-by-id))))
+
+(if-bb
```
oliyh, line 138: If I understand correctly, you let `lein test` start the server and then this will shell out to bb to run the bb tests while the server is still running - is that right?

And the tests for this module are run not by `lein sub` from the parent project, but instead by the extra addition to the circle config?

core/test/martian/test_helpers.cljc
borkdude: This is an issue specifically with bb:

```
user=>  (defmacro yaml [s] (doto (clj-yaml.core/parse-string s) prn))
#'user/yaml
user=>  (yaml "{a: [1 2 3]}")
#ordered/map ([:a ("1 2 3")])
class java.lang.String cannot be cast to class clojure.lang.IFn (java.lang.String is in module java.base of loader 'bootstrap'; clojure.lang.IFn is in unnamed module of loader 'app')
```

In clojure you can reproduce a similar thing with:

```
user=> (defmacro my-map [x] {:a (map inc [1 2 3])})
#'user/my-map
user=> (my-map 1)
Execution error (ClassCastException) at user/eval172 (REPL:1).
class java.lang.Long cannot be cast to class clojure.lang.IFn (java.lang.Long is in module java.base of loader 'bootstrap'; clojure.lang.IFn is in unnamed module of loader 'app')
```

Adding a quote in front of it fixes it:

```
user=> (defmacro my-map [x] (list 'quote {:a (map inc [1 2 3])}))
#'user/my-map
user=> (my-map 1)
{:a (2 3 4)}
```

It can be argued that the behavior in bb is a bug since clojure didn't have this problem, but adding a `'quote` doesn't hurt either way.

babashka-http-client/test/martian/babashka/http_client_test.clj
borkdude: > is that right?

This is right.

The tests here should be run like any of the other modules.

> instead by the extra addition to the circle config 

This was only to run tests for the core module.

borkdude: So I think I forgot to configure something for `lein sub`?

borkdude: For the record, I have no idea what lein sub is

oliyh: I actually mean `lein-modules` (I used to use `lein-sub` but not anymore). lein-modules is a plugin for lein that lets you have multiple child lein projects which can share some configuration around versioning and running tasks - e.g. running `lein install` from the top directory will run `lein install` on each child project (in the right order to satisfy dependencies).
https://github.com/jcrossley3/lein-modules

But in this case this new module already supports it with the inclusion of `:plugins [[lein-modules "0.3.11"]]` in the project.clj so I think we are all good here.

borkdude:
@oliyh OK, processed your feedback now.

core/src/martian/encoders.cljc
```diff
@@ -1,6 +1,7 @@
 (ns martian.encoders
   (:require [clojure.string :as string]
-            [linked.core :as linked]
+            #_[linked.core :as linked]
```
oliyh, line 3: This line can be removed

core/src/martian/encoders.cljc
```diff
@@ -36,6 +37,9 @@
   ([] (default-encoders keyword))
   ([key-fn]
    (merge
+    #?(:bb
+       {"application/transit+json"    {:encode #(transit-encode % :json)
+                                       :decode #(transit-decode % :json)}})
     #?(:clj
        {"application/transit+msgpack" {:encode #(transit-encode % :msgpack)
```
oliyh, line 44: We still need transit+json support for `clj` (but it needs to come after transit+msgpack)

core/src/martian/encoders.cljc
```diff
@@ -36,6 +36,9 @@
   ([] (default-encoders keyword))
   ([key-fn]
    (merge
+    #?(:bb
```
oliyh, line 39: For clj, we should support and prefer (in this order)
1. transit+msgpack
2. transit+json
3. edn
4. json

For bb and cljs it should just be 2, 3, 4 as they don't support msgpack.

So you can probably put msgpack for `:clj` only at the top, and merge that with a common list of 2, 3 and 4 to make it a bit more concise, i.e.

```clj
(merge
  #?(:clj {"application/transit+msgpack" ...})
  {"application/transit+json" ...
   "application/edn" ...
   "application/json" ...})
```

borkdude: This doesn't work for bb since bb will take the `:clj` branch if it's before the `:bb` branch.

oliyh:
Thanks @borkdude I thought there was an opportunity to be more concise but as you say bb will take the clj option, and I also noticed that the cljs branch wasn't identical either, so I think what you have is correct.

borkdude:
Are the failing tests of any concern?

oliyh:
The failing tests are because I lost https://pedestal-api.herokuapp.com when heroku effectively ended their free tier, and I haven't found time to deploy it anywhere else yet, so nothing to worry about here.

oliyh:
Thanks for this!

borkdude:
Amazing, thank YOU!

PR #173, opened 2023-04-07
removing Clojars coordinates notice from README

The-Alchemist:
The note is from 2021-07-13, almost two years ago.

It makes it seem like the project is no longer active because old documentation, which is inaccurate. ;)


oliyh:
Thanks!

Issue #111, opened 2021-04-22
Issue with required parameters


evilsneer:
Hi!

I use swagger as below, and schema, which martian generates contains keys of data dict as required, but it is not. So queries "fails" on client side.

 Swagger-ui handles this case (no red stars above fields in model view).

```

{
    "swagger": "2.0",
    "x-id": [
        "mango.routes.services/api"
    ],
    "info": {
        "title": "my-api",
        "description": "https://cljdoc.org/d/metosin/reitit"
    },
    "paths": {
        "/api/datasets/phones": {
            "post": {
                "responses": {
                    "default": {
                        "description": ""
                    }
                },
                "produces": [
                    "application/json",
                    "application/transit+msgpack",
                    "application/transit+json",
                    "application/edn"
                ],
                "consumes": [
                    "application/json",
                    "application/transit+msgpack",
                    "application/transit+json",
                    "application/edn"
                ],
                "parameters": [
                    {
                        "in": "body",
                        "name": "body",
                        "description": "",
                        "required": true,
                        "schema": {
                            "type": "object",
                            "properties": {
                                "phone": {
                                    "type": "object",
                                    "properties": {
                                        "operation": {
                                            "title": "Operation",
                                            "description": "operations on field",
                                            "enum": [
                                                "=",
                                                "contains",
                                                "startswith",
                                                "like"
                                            ],
                                            "default": "=",
                                            "type": "string"
                                        },
                                        "value": {
                                            "type": "string",
                                            "x-nullable": true
                                        }
                                    },
                                    "required": [
                                        "operation",
                                        "value"
                                    ]
                                },
                                "gender": {
                                    "type": "object",
                                    "properties": {
                                        "operation": {
                                            "title": "Operation",
                                            "description": "operations on field",
                                            "enum": [
                                                "=",
                                                "startswith",
                                                "like",
                                                "contains"
                                            ],
                                            "default": "=",
                                            "type": "string"
                                        },
                                        "value": {
                                            "type": "string",
                                            "x-nullable": true
                                        }
                                    },
                                    "required": [
                                        "operation",
                                        "value"
                                    ]
                                },
                                "age": {
                                    "type": "object",
                                    "properties": {
                                        "operation": {
                                            "title": "Operation",
                                            "description": "operations on field",
                                            "enum": [
                                                "=",
                                                "contains",
                                                "startswith",
                                                "like"
                                            ],
                                            "default": "=",
                                            "type": "string"
                                        },
                                        "value": {
[...]

PR #177, opened 2023-05-03
Fix requirement of nested objects

gabrielgiussi:
### The problem

Nested objects mark as required in a swagger Schema Object are translated to optional keys.
See the test `require-nested-objects` for an real use case obtained from the k8s api with chaos-mesh installed.

### A possible solution

The solution implemented in this PR is adding a `denormalised` attribute to prevent being denormalised again which seems to be what is causing the issue. 
Perhaps someone with a better understanding of the codebase can provide a more principled solution.
The main contribution is the test showing which scenario is not working properly.

### DIFF
--- a/core/project.clj
+++ b/core/project.clj
@@ -27,7 +27,8 @@
                                   [com.bhauman/figwheel-main "0.2.13"]
                                   [org.clojure/tools.reader "1.3.5"]
                                   [cider/piggieback "0.5.2"]
-                                  [org.clojure/tools.nrepl "0.2.13"]]
+                                  [org.clojure/tools.nrepl "0.2.13"]
+                                  [nubank/matcher-combinators "3.8.5"]]
                    :repl-options {:nrepl-middleware [cider.piggieback/wrap-cljs-repl]}}}
   :aliases {"fig"       ["trampoline" "run" "-m" "figwheel.main"]
             "fig:build" ["trampoline" "run" "-m" "figwheel.main" "-b" "dev" "-r"]
--- a/core/test/martian/schema_test.cljc
+++ b/core/test/martian/schema_test.cljc
@@ -1,5 +1,6 @@
 (ns martian.schema-test
   (:require [martian.schema :as schema]
+            [matcher-combinators.test]
             [schema.core :as s]
             [schema-tools.core :as st]
             #?(:clj [clojure.test :refer [deftest testing is]]
@@ -135,7 +136,7 @@
                                       {:name "Brachiosaurus"
                                        :address {:city nil}}
                                       nil
-                                       true))))
+                                      true))))
 
          (testing "adds missing keys"
            (is (= {:name "Brachiosaurus"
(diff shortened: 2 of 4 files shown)


core/project.clj
```diff
@@ -27,7 +27,8 @@
                                   [com.bhauman/figwheel-main "0.2.13"]
                                   [org.clojure/tools.reader "1.3.5"]
                                   [cider/piggieback "0.5.2"]
-                                  [org.clojure/tools.nrepl "0.2.13"]]
+                                  [org.clojure/tools.nrepl "0.2.13"]
+                                  [nubank/matcher-combinators "3.8.5"]]
```
gabrielgiussi, line 31: I've added this library that helps to write more concise assertions (and also provides a colored diff when assertions fail to spot easily the difference btw expected and actual), but let me know if you prefer me to remove it.

core/test/martian/schema_test.cljc
```diff
@@ -394,3 +395,65 @@
 
     (is (= (s/maybe {(s/optional-key :b) (s/maybe {(s/optional-key :a) s/Any})})
            schema))))
+
+
+(def definition {:org.chaos-mesh.v1alpha1.NetworkChaos {:description "NetworkChaos is the Schema for the networkchaos API",
```
gabrielgiussi, line 400: I could use a json like swagger-issue-111.json to improve readability if you prefer

oliyh:
Hi,

Thank you for the contribution, it's much appreciated. I will try to find time soon to go through it.

I am in favour of improved test output but unfortunately this library doesn't seem to be compatible with babashka so the babashka build is failing, see CircleCI. Can this library just be used on the JVM?

Thanks

gabrielgiussi:
I'm not sure, I've tried to follow [these instructions](https://github.com/babashka/babashka/blob/master/doc/dev.md#tests-for-libraries) but the script is failing with
> Error building classpath. Manifest type not detected when finding deps for babashka/babashka.core in coordinate #:local{:root "/Users/gabriel.giussi/repos/babashka/babashka.core"}

So I'll just remove matcher-combinators.

gabrielgiussi:
It seems we also have to add matcher-combinators in the bb file since it doesn't know about the dev alias defined in project.clj.
BTW babashka itself uses matcher-combinators https://github.com/babashka/babashka/blame/e84eb29fa16de78625309d31a853435218bd35fb/deps.edn#L185

oliyh:
Hi,

Thanks for this, I've had a look and found that the pre-emptive denormalisation of object graphs introduced in #111 has resulted in this behaviour. It is unneccessary, because the object graph is walked anyway, but it did have the benefit of disambiguating the `required` field which differs between leaves and objects.

I've done some work on a different approach which passes all tests except for the one introduced for #111, but I think it's nearly there.

https://github.com/oliyh/martian/compare/master...fix-required-nested-objects

gabrielgiussi:
Great, thanks for taking a look to this. Feel free to close this PR if you are going to be working on that branch.

oliyh:
I'll leave it open as a reminder to me if you don't mind, and I'll update you here when I've got something into the main branch.

Cheers

oliyh:
Hi,

Apologies for the delayed response. I have decided that the test failed because we have in fact improved the behaviour - it had an optional key which was itself an object, which had two required parameters - so the object could not be a maybe. Merged in #197 

Thank you for your work! I will publish a snapshot to clojars and a release soon.

Cheers

oliyh:
Released in `0.1.27-SNAPSHOT`

PR #191, opened 2023-11-05
[doc] Mark babashka http client as JVM compatible

lispyclouds:
The Babashka http client also runs on the JVM and this marks it as such.


oliyh:
Thanks!

PR #193, opened 2023-11-11
OpenAPI's integer can have default value `inf`

robertluo:
https://github.com/openai/openai-openapi/blob/5c1857ea865e74e45e3b12064e0cc2396ef64be1/openapi.yaml#L6012C2-L6012C2

The default value for the integer is "inf".


core/src/martian/schema.cljc
```diff
@@ -98,7 +98,13 @@
 
 (defn wrap-default [{:keys [default]} schema]
   (if (some? default)
-    (st/default schema default)
+    (let [default
+          ;patch for `inf` as a default value
+          (if (and (= schema s/Int) (= default "inf"))
+            #?(:clj Long/MAX_VALUE
+               :cljs (Math/pow 10 1000))
```
oliyh, line 105: Is this the best that can be done on cljs?

robertluo: No. In fact, I realized the cljs test failed, but I have trouble to run the tests locally. Do you know how do I specify the correct Chrome location to run the figwheel tests?

oliyh: You can set the path to chrome here: https://github.com/oliyh/martian/blob/master/core/test.cljs.edn#L5

robertluo:
I have used my fork with this patch for the latest OpenAI API for a few days, and it looks good.

oliyh:
Thank you for this!

oliyh:
It's now in the latest snapshot `0.1.26-SNAPSHOT`

Issue #195, opened 2024-01-23
$ref is not working properly for swagger parameters


RafaeLeal:
Hi! I think I've discovered a bug related to `$ref` usage on parameters for swagger. I've tried to reproduce it using OpenAPI v3, but it seems it's working properly there.

Consider the schema:
```yaml
swagger: "2.0"
info:
  title: "Users API"
  version: "1.0"
definitions:
  userIdParam:
    name: id
    in: query
    description: user's id
    required: true
    type: string
paths:
  /users:
    get:
      operationId: getUsers
      parameters:
        - $ref: "#/definitions/userIdParam"
      responses:
        '200':
          description: "OK"
```

It's not producing the query schema correctly:
```clj
  (let [client (martian-httpkit/bootstrap-openapi "swagger-example.yaml")]
    (martian/explore client :get-users))
=> {:summary nil, :parameters {}, :returns {200 Any}}
```

oliyh:
Closing as fixed in #196

PR #196, opened 2024-01-23
Fix Reference Object usage on parameters for swagger spec

RafaeLeal:
As I commented in the issue
https://github.com/oliyh/martian/issues/195

This is not working for swagger, but it is for OpenAPI v3. I tried to refactor the code to use the same logic, adding tests for it.

### DIFF
--- a/core/src/martian/schema.cljc
+++ b/core/src/martian/schema.cljc
@@ -75,6 +75,11 @@
   (let [[_ category k] (re-find #"#/(definitions|parameters)/(.*)" ref)]
     (get-in ref-lookup [(keyword category) (keyword k)])))
 
+(defn resolve-ref-param [ref-lookup param]
+  (if-let [ref (:$ref param)]
+    (resolve-ref ref-lookup ref)
+    param))
+
 (def URI
   #?(:clj java.net.URI
      :cljs goog.Uri))
--- a/core/src/martian/schema.cljc
+++ b/core/src/martian/schema.cljc
@@ -5,6 +5,7 @@
             [schema.coerce :as sc]
             [schema-tools.core :as st]
             [schema-tools.coerce :as stc]
+            [clojure.string :as string]
             [martian.parameter-aliases :refer [unalias-data]])
   #?(:clj (:import [schema.core AnythingSchema Maybe EnumSchema EqSchema])))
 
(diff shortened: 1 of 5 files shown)


core/test/martian/swagger_test.cljc
```diff
@@ -18,6 +18,28 @@
     (is (= {:id s/Str} (:path-schema get-handler)))
     (is (= nil (:query-schema get-handler)))))
 
+(deftest parameter-schemas-ref-test
```
RafaeLeal, line 21: This is the main scenario of the bug

core/src/martian/schema.cljc
```diff
@@ -71,9 +72,31 @@
...
+(defn lookup-ref
+  [ref ref-lookup]
+  (if (string/starts-with? ref "#/")
+    (or (get-in ref-lookup (drop 1 (map keyword (string/split ref #"/"))))
+        (throw (ex-info "Cannot find reference"
+                        {:reference ref})))
+    (throw (ex-info "Non-local references are not supported yet. References should start ref with '#/'"
+                    {:reference ref}))))
+
+(defn resolve-ref-object
```
RafaeLeal, line 84: this function follows through when you have `$ref` pointing to other `$ref`. This is possible following OpenAPI v3, but not Swagger. I believe this scenario was not covered previously, so I added it here.

core/test/martian/schema_test.cljc
```diff
@@ -400,3 +401,34 @@
...
     (is (= (s/maybe {(s/optional-key :b) (s/maybe {(s/optional-key :a) s/Any})})
            schema))))
+
+
+
+(deftest resolve-ref-object-test
+  (let [params       {:idParam {:name "id" :in "query"}}
+        swagger-spec {:definitions params}
+        openapi-spec {:components {:parameters params}}
+        #?(:cljs ExceptionInfo) #?(:cljs js/Error)]
```
RafaeLeal, line 411: This is a hack to work with ClojureScript..
we might want to change it to some test-helper like this:
https://gist.github.com/metametadata/69cae3488d891f5e23d9

oliyh: I tend to just do this at the top level:
```
#?(:cljs (def ExceptionInfo js/Error))
```

oliyh:
This is great, thank you!

RafaeLeal:
Thanks for the merge!
Can you generate a version for it? 
I'm using it on [k8s-api](https://github.com/nubank/k8s-api) and the newer versions of Kubernetes needs it

oliyh:
Hi @RafaeLeal I have release `0.1.26`

RafaeLeal:
@oliyh Thanks!

PR #201, opened 2024-04-27
Improve error handling

velios:


velios:
Hello. First of all, thx for exellent lib.

I happened to be parsing someone else's API and came across this probably incorrect entry. The problem is not that it is incorrect, but that an uninformative error was received.

Problem json part
```json
"responses": {
          "200": {
            "description": "successful operation",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "properties": {
                    "profiles": {
                      "type": "array",
                      "items": {
                        "type": "object",
                        "properties": {
                          "id": "string",
                          "depth": "number",
                          "site-id": "string"
                        },
                        "required": [
                          "id",
                          "depth"
                        ],
                        "example": "some example"
                      }
                    }
                  },
                 ...
```

When `(keys schema)` call with `"string"` value in `"id": "string"` string i get

```clojure
#error {
 :cause nil
 :via
 [{:type java.lang.ClassCastException
   :message nil}]
 :trace
 []}
```

I would like to somehow improve the output for such cases, since I came across it. The wording does not stand up to criticism, I understand, so I hope that you will find the time to correct it for an understandable error.

oliyh:
Hello!

Thank you for your contribution. I agree the existing message is not very helpful. I will hopefully find time to look at this more closely soon.

Cheers

velios:
What’s worse is that with some errors I can get an instance with partially loaded apis. Since in this case I will not receive the correct notification of problems, I will assume that my entire API has loaded.

Issue #181, opened 2023-07-22
VCR playback should terminate if it has a response


oliyh:
VCR playback should terminate if it has a response (whatever it is). This will prevent a real perform-request interceptor from being invoked even if the playback response should have been used.

lvh:
FYI I don't think the behavior introduced in that commit is correct. I successfully bisected a bug down to it: https://gist.github.com/lvh/9e765f8999cb6d7f5a93a5abbd994588

I'm not 100% sure I identified the bug correctly, but, let's start with observed behavior:

* When returning a VCR response, `:body` is now a String instead of the expected parsed map (the underlying API returns JSON)
* When not returning a VCR response, the code works fine (meaning in production for example)
* This behavior is consistent with the interceptor chain ending _entirely_ instead of just terminating the :enter stage; meaning, the :leave stage is not started (which is where the parsing code runs)
* That's not how tripod's `terminate` is supposed to work, so I think this is because `remove-stack` deletes the interceptors  and then terminates
* I think the correct behavior is to `terminate` but not `remove-stack`

lvh:
I'm also not sure I understand the use case for this behavior: isn't the VCR replaying interceptor supposed to replace the real-HTTP-request-making one, so there's no chance of a real HTTP request being made anyway? Is this change intended to do anything if the user is doing that correctly?

FWIW I tried to replace remove-stack with terminate and that did ostensibly fix the issue. I'll create a PR.

lvh:
I think I figured out why this failure doesn't occur in the test.  Unfortunately, I was unable to debug this particularly closely because I couldn't start a REPL; the issue looks identical to https://github.com/oliyh/martian/issues/85 even though it is running in a subdirectory:

```
(cd /Users/lvh/Projects/martian/vcr; lein update-in :dependencies conj '[nrepl,"1.1.1"]' -- update-in :plugins conj '[cider/cider-nrepl,"0.47.1"]' -- update-in '[:repl-options,:nrepl-middleware]' conj '["cider.nrepl/cider-middleware"]' -- with-profile +dev repl :headless)
Error encountered performing task 'repl' with profile(s): 'base,system,user,provided-martian-vcr,dev-martian-vcr'
java.lang.IllegalArgumentException: Bad artifact coordinates com.github.oliyh:martian:jar::version, expected format is <groupId>:<artifactId>[:<extension>[:<classifier>]]:<version>
```

[...]

PR #204, opened 2024-06-09
Terminate the stack but don't remove the rest when using VCR

lvh:
References #181.

`remove-stack` prevents all further interceptors from running. It uses `terminate` under the hood. Critical difference: `terminate` just means we'd move to the second phase (`:leave`), ensuring things like e.g. response body encoding still happens.

### DIFF
--- a/vcr/src/martian/vcr.cljc
+++ b/vcr/src/martian/vcr.cljc
@@ -2,8 +2,8 @@
   (:require #?@(:clj [[clojure.java.io :as io]
                       [clojure.edn :as edn]
                       [fipp.clojure :as fipp]
-                      [martian.interceptors :refer [remove-stack]]]
-                :cljs [[martian.interceptors :refer [remove-stack]]])))
+                      [tripod.context :as tc]]
+                :cljs [[tripod.context :as tc]])))
 
 (defmulti persist-response! (fn [opts _ctx] (get-in opts [:store :kind])))
 (defmulti load-response (fn [opts _ctx] (get-in opts [:store :kind])))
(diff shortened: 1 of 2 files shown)


oliyh:
This looks good, I am very grateful for your efforts here. I'm concerned that no tests failed! I would like to understand that first before merging this.

lvh:
Sure! I'm not sure if you saw my comments on #181 already when writing that. My theory for why the tests didn't fail is that the dummy response in the test had an already-parsed body, but a real HTTP client would return a String. Normally, the :leave stage would parse it for you. Here that wasn't happening but nobody noticed because the output didn't need parsing. Hence my semantic confusion comment from #181: I think the test had a false negative because the dummy response is incorrectly being treated as both an HTTP response hot off the client, and an otherwise already cleaned up response-for.

I think a good test would be for the dummy response to return a String but the test to validate that it is no longer a String, hence validating :leave ran.

oliyh:
Yes, that would be a good test to have - thank you for explaining the current behaviour

lvh:
Well, I'm glad I wrote the test because now I'm not sure I'm doing it correctly. I added a commit, but the tests don't pass :|

Could you check me on the test behavior and confirm that it's asking for the correct behavior?

lvh:
OK, I think I figured out why the test fails, and good news: it's a spurious failure.

1. In at least one case the test was just wrong
2. In at least one case the default coercion wasn't working because it expected response headers to be camel-cased but they're actually TitleCase--the logic that ordinarily makes those consistent wasn't loaded in the tests.

FWIW this was a lot harder than it should've been because I'm unable to get a REPL working.

lvh:
Hooray, the tests pass now! I'm told this means I'm doing test-driven development correctly. It certainly could not possibly mean I screwed the test up before :)

@oliyh: I appreciate you just did a release but given that we seem to be in agreement (or at least hopefully will be) that:

1. The old behavior is wrong
2. It's wrong in a way that caused a real problem in a real project
3. The test now convincingly has the bug dead to rights

... I'd be very appreciative of a release since that would fix my project :)

oliyh:
Released in `0.1.30` :partying_face:

lvh:
Wooo, thank you very much!

PR #205, opened 2024-06-09
Ensure graceful failure for implementation-specific objects that vcr writes but can't read back

### LINKED ISSUE #203: martian-hato puts implementation objects in the request that martian-vcr can write but not read back
When using martian-hato with martian-vcr (and pretty printing on, FWIW, see vcr opts below):

```clojure
(def vcr-opts
  {:store
   {:kind :file
    :root-dir "test-resources/REDACTED-vcr"
    :pprint? true}
   :on-missing-response :throw-error})
```

This is used like this:

```clojure
(defn recording-bootstrap
  "A variant of `bootstrap` that saves responses with vcr."
  []
  (let [record (martian-vcr/record vcr-opts)]
    (li/bootstrap
     {:opts-hook
      (fn [opts]
        (li/inject opts record :before (:name martian-http/perform-request)))})))

(defn playbacking-bootstrap
  "A variant of `bootstrap` that returns previously-saved results using vcr."
  []
  (let [playback (martian-vcr/playback vcr-opts)]
    (li/bootstrap
     {:opts-hook
      (fn [opts]
        (li/inject opts playback :replace (:name martian-http/perform-request)))})))
```

Just to clarify the require: 1[martian.hato :as martian-http]`

The `li/bootstrap` referred to is:

```clojure
(defn bootstrap
  "A custom version of [[martian.hato/bootstrap-openapi]] that adds an
   authentication ([[add-authn]]) via interceptor.

  Optionally takes a map with the following optional arguments:

  * `opts-hook` gets called with the martian bootstrap opts right before they're
  passed to bootstrap, allowing you to customize interceptors (for example). If
  you're using this there's a good chance you want to look at [[inject]]."
  ([] (bootstrap nil))
  ([{:keys [opts-hook] :or {opts-hook identity}}]
   (let [definition (load-api-definition!)
         base-url (openapi/base-url openapi-url nil definition)
         opts (-> martian-http/default-opts
                  (inject add-authn :before (:name martian-http/perform-request))
                  opts-hook)]
     (martian/bootstrap-openapi base-url definition opts))))
```

Free free to use that a feature suggestion if it's helpful :) Happy to provide `inject` too, but I'm sure you can guess the implementation: it's essentially a version of `inject` that works on `opts` instead of on the list of interceptors directly:

```clojure
(defn inject
  "Like [[int/inject]] for interceptors, but acts on the martian `opts` map that
  _contains_ the interceptors."
  [opts & args]
  (apply update opts :interceptors int/inject args))
```

The resulting saved EDN file looks like this:

```edn
{:request-time 756,
 :request {:user-info nil,
           :as :text,
           :headers {"Accept" "application/json"},
           :url "https://REDACTED",
           :http-request #object[jdk.internal.net.http.HttpRequestImpl "0x2befd612" "REDACTED"]
           ;; ... irrelevant keys redacted
           :request-method :get},
 :http-client #object[jdk.internal.net.http.HttpClientFacade
                      "0x2e234aeb"
                      "jdk.internal.net.http.HttpClientImpl@1a162bb8(7)"],
 :headers {"content-encoding" "gzip",
           ;; irrelevant keys redacted
           "report-to" "REDACTED"},
 :status 200,
 :content-type :application/json,
 :uri "REDACTED",
 :content-type-params {},
 :version :http-1.1,
 :body "REDACTED"}
```

The write succeeds but the read never will because the `#object` reader tag can't actually be deserialized.

I think I can work around this with a carefully placed interceptor to go along with the recording one.

lvh:
Fixes #203

### DIFF
--- a/vcr/src/martian/vcr.cljc
+++ b/vcr/src/martian/vcr.cljc
@@ -49,10 +49,19 @@
                     (pr-str response))))))
 
 #?(:clj
+   (defn ^:private unreadable-entry-handler
+     [tag form]
+     {:tag tag
+      :form form
+      :error "Unreadable entry in from vcr"}))
+
+#?(:clj
    (defmethod load-response :file [{:keys [extra-requests] :as opts} ctx]
      (let [file (response-file opts ctx)]
        (if (.exists file)
-         (edn/read-string (slurp file))
+         (edn/read-string
+          {:default unreadable-entry-handler}
+          (slurp file))
          (some->
           (condp = extra-requests
             :repeat-last (last-response opts ctx)


lvh:
Happy to try and write a test for this if you have some guidance on how to do it. Also would prefer to get #204 merged first since the tests will almost certainly conflict.

Proposal: just have the underlying dummy responder in the test return a function (`(constantly 'xyzzy)`) works).

oliyh:
Hi thank you for this, it's a nice general solution.

Looks like no tests have changed - either here on in the other PR - we can merge this one now?

Cheers

lvh:
I don't think there's a reason not to, no: only reason would be if you thought there was good reason the behavior was brittle and as such it's important to write a specific test.

oliyh:
Thank you for this!

Issue #215, opened 2025-01-05
Missing-key / disallowed key on path-params when setting query params


licht1stein:
Here's my example:

```clojure
 (def m (martian/bootstrap
          "http://example.com"
          [{:query-schema {:version (st/default s/Int 70)},
            :path-schema {:id s/Str}
            :route-name :test/route
            :method :get,
            :path-parts ["/part/" :id]}]
          {:use-defaults? true}))

  (martian/request-for m :test/route {:id "honestly-i-am-a-string"})
  ;; Value cannot be coerced to match schema: {:id disallowed-key}

  (martian/request-for m :test/route {})
  ;; Could not coerce value to schema: {:id missing-required-key}
```

I tried working from the readme test example, and figured out that adding a `:query-schema` map breaks it:

```clojure
(let [m (martian/bootstrap "https://api.org"
                   [{:route-name :load-pet
                     :path-parts ["/pets/" :id]
                     :query-schema {:version (st/default s/Int 70)}
                     :method :get
                     :path-schema {:id s/Str}}]
                   {:use-defaults? true})]
    (martian/request-for m :load-pet {:id "1"}))
;; Value cannot be coerced to match schema: {:id disallowed-key}
```

licht1stein:
Moreover, when removing the defaults part everything works:

```clojure
(let [ m (martian/bootstrap
         "http://example.com"
         [{:query-schema {:version s/Int},
           :path-schema {:id s/Any}
           :route-name :test/route
           :method :get,
           :path-parts ["/part/" :id]}]
         )]

    (martian/request-for m :test/route {:id "honestly-i-am-a-string" :version 1}))
;; => {:method :get, :url "http://example.com/part/honestly-i-am-a-string", :query-params {:version 1}}
```

oliyh:
Hi,

Thanks for the explanation and example. This is very odd, I can't explain this behaviour. I will have a look into this when I have a chance.

Cheers

marksto:
Hi guys,

I confirm that the behaviour exists, does indeed depend on the `:use-defaults?` presence, and kinda makes sense.

If I override the `martian.interceptors/coerce-data` fn like this:

```clojure
(defn coerce-data [{:keys [parameter-aliases] :as handler} schema-key params opts]
  (prn "schema-key ->" schema-key)
  (prn "params     ->" params)
  (let [schema        (get handler schema-key)
        param-aliases (get parameter-aliases schema-key)
        use-defaults? (:use-defaults? opts)
[...]

PR #219, opened 2025-05-06
Fix how default values interplay with map schemas coercion

marksto:
This fixes up the issue with [Missing-key / disallowed key on path-params when setting defaults in query-schema #215](https://github.com/oliyh/martian/issues/215).

```clojure
(require
  '[martian.core :as martian]
  '[spec-tools.core :as st])

(def m (martian/bootstrap
         "http://example.com"
         [{:query-schema {:version (st/default s/Int 70)},
           :path-schema  {:id s/Str}
           :route-name   :test/route
           :method       :get,
           :path-parts   ["/part/" :id]}]
         {:use-defaults? true}))
=> #'user/m

(martian/request-for m :test/route {:id "honestly-i-am-a-string"})
=> {:method :get, :url "http://example.com/part/honestly-i-am-a-string", :query-params {:version 70}}

(martian/request-for m :test/route {})
Execution error (ExceptionInfo) at schema-tools.coerce/coerce-or-error! (coerce.cljc:24).
Could not coerce value to schema: {:id missing-required-key}
```

All existing tests pass locally (Clojure). I've also added a new one for this issue, `martian.core-test/use-defaults-test`. It goes with all 4/4 cases that were mentioned in the issue, i.e. it provides a complete coverage:

```clojure
(require
  '[martian.schema :as schema]
  '[spec-tools.core :as st])

(schema/coerce-data
  {:version (st/default s/Int 70)}
  {}
  nil true)
=> {:version 70}

(schema/coerce-data
  {:version (st/default s/Int 70)}
  {:version 100}
  nil true)
=> {:version 100}

;; This produces "Value cannot be coerced to match schema: {:id disallowed-key}" w/o the fix.
(schema/coerce-data
  {:version (st/default s/Int 70)}
  {:id "id"}
  nil true)
=> {:version 70}

;; This produces "Value cannot be coerced to match schema: {:id disallowed-key}" w/o the fix.
(schema/coerce-data
  {:version (st/default s/Int 70)}
  {:id "id" :version 100}
  nil true)
=> {:version 100}
```

### DIFF
--- a/core/src/martian/interceptors.cljc
+++ b/core/src/martian/interceptors.cljc
@@ -36,7 +36,12 @@
             (update ctx :request create-only :url (url-for (:route-name handler) params)))})
 
 (defn coerce-data [{:keys [parameter-aliases] :as handler} schema-key params opts]
-  (schema/coerce-data (get handler schema-key) params (get parameter-aliases schema-key) (:use-defaults? opts)))
+  (let [schema (get handler schema-key)
+        ;; A map schema needs to be open prior to params coercion (#215)
+        schema (if (and (map? schema) (nil? (s/find-extra-keys-schema schema)))
+                 (assoc schema s/Keyword s/Any)
+                 schema)]
+    (schema/coerce-data schema params (get parameter-aliases schema-key) (:use-defaults? opts))))
 
 (def keywordize-params
   {:name ::keywordize-params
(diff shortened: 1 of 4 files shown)


marksto:
@oliyh Hi Oliver! The fix is ready for your review. It would be nice to incorporate this into the next Martian release.

oliyh:
Hi @marksto ,

Wow, well done - I really appreciate the perseverance with this tricky issue. It's been a long time since I looked at this code so I can't quite remember how it all plays together, but your tests are very clear.

Thanks for this

Issue #220, opened 2025-05-07
Way too lax coercions hinder proper validation / Set coercion matchers via `:opts` maybe?


marksto:
Hi Oliver!

I just noticed how by default it does coerce boolean values if a string gets there:

```clojure
(schema/coerce-data
  {:a-bool-maaan s/Bool}
  {:a-bool-maaan "WAT?"})
=> {:a-bool-maaan false}
```

This is due to the fact that a `schema.coerce/string-coercion-matcher` kicks in, which itself has these boolean coercion rules inside:

```clojure
s/Bool string->boolean
Boolean string->boolean
```

where `string->boolean` fn "returns true for strings that are equal, ignoring case, to the string 'true'".

This is not what I expect (in fact, this is what I want to avoid, as if I were using the default `schema-tools.coerce/json-coercion-matcher`) in my particular case. Here I expect a coercion / validation error, e.g.:

```clojure
(schema/coerce-data
  {:a-bool-maaan s/Bool}
  {:a-bool-maaan "WAT?"})
Execution error (ExceptionInfo) at schema-tools.coerce/coerce-or-error! (coerce.cljc:24).
Could not coerce value to schema: {:a-bool-maaan (not (instance? java.lang.Boolean "WAT?"))}
```

At the moment I can only have a simple workaround for this globally, by altering some vars' roots, which is not cool.

**Can we explicitly expose these coercion matchers so that they become part of the Martian's API?**

There are already `:opts` and even `:use-defaults?` that hit the right spot (the `coerce-data` fn).
Maybe we can add one more global option, say, `:coercion-matchers`?
And I also believe this specific feature should not be handler-specific. WDYT?

Cheers,
Mark

PR #221, opened 2025-05-08
Support setting coercion matchers via Martian `:opts`

marksto:
@oliyh Hello again, Oliver!

This PR addresses a proposal made in #220.

The main part is essentially non-breaking (passing things through existing `:opts` to override the `default-coercion-matcher`), but I've also made a few minor relevant improvements here and there — improved readability/maintainability, exposed some useful schema internals (private -> public), reused `schema-tools` fns, and reordered `core` ns imports.

Thanks for such a neat tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Martian
+
 Calling HTTP endpoints can be complicated. You have to construct the right URL with the right route parameters, remember
 what the query parameters are, what method to use, how to encode the body and many other things that leak into your codebase.
 
@@ -56,6 +57,7 @@ Testing and other interop libraries:
 
 
 ## Features
+
 - Bootstrap an instance from just a OpenAPI/Swagger url, a local definition file or provide your own API mapping
 - Modular with support for many HTTP client libraries (see table above)
 - Build urls and request maps from code or generate and perform the request, returning the response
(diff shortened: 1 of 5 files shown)


oliyh:
Hi,

This looks good to me, thank you for all your efforts - definitely some of the trickiest parts of the codebase that you've waded into, and done a good job!

I was thinking that I was sure I'd already done something like this, and then realised it was the en/decoding extension point which is analogous. I thought that perhaps a solution to this could be similar, but of course the coercion of data is spread out across multiple interceptors (perhaps needlessly?) and things like the aliasing further complicate matters.

So this is a good way of injecting this behaviour into the current stack. If I were ever to rewrite martian I think consolidating the schema coercion to a single interceptor, and making the schema library more pluggable (so that using malli would be possible) would probably be top of the list.

I will merge this now and try to get a release out (I have also been having a lot of issues with the lein submodule plugin which no longer works...)

oliyh:
Released in `0.1.31` :tada:

marksto:
> If I were ever to rewrite martian I think consolidating the schema coercion to a single interceptor, and making the schema library more pluggable (so that using malli would be possible) would probably be top of the list.

Yeah, I totally second this plan. Having Malli instead of Schema would be awesome!

PR #225, opened 2025-05-16
Add support for 'multipart/form-data' content-type (JVM/Babashka)

marksto:
@oliyh  Hi Oliver!

Here's another itch I had to scratch recently. 😅 

This one is tiny and straightforward — it adds support for `:multipart` requests for all target JVM/Babashka HTTP clients that support this content-type (basically, all except `clj-http-lite`).

The change is backward compatible:
- I didn't drop the original `default-encode-body` fn — it's now effectively an alias for the `default-encode-response`
- As I have not changed the main path in the work of the `encode-response` (ex-`encode-body`)
- Custom end user interceptors will still override the default ones with a user-specific "multipart/form-data" encoder

Hope this one as well helps! Please, tell me if I am missing something important and not that obvious.

Cheers,
Mark

### DIFF
--- a/core/src/martian/interceptors.cljc
+++ b/core/src/martian/interceptors.cljc
@@ -1,5 +1,6 @@
 (ns martian.interceptors
   (:require [martian.schema :as schema]
+            [clojure.set :as set]
             [clojure.walk :refer [keywordize-keys stringify-keys]]
             [tripod.context :as tc]
             [schema.core :as s]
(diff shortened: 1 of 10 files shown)


core/src/martian/interceptors.cljc
```diff
@@ -82,19 +83,27 @@
...
               ctx))})
 
-(defn encode-body [encoders]
-  {:name ::encode-body
+(defn encode-request [encoders]
+  {:name ::encode-request
    :encodes (keys encoders)
    :enter (fn [{:keys [request handler] :as ctx}]
-            (let [content-type (and (:body request)
+            (let [has-body? (get-in ctx [:request :body])
```
oliyh, line 90:
```suggestion
            (let [has-body? (:body request)
```

oliyh:
Thanks!

PR #230, opened 2025-06-21
Refine multipart requests handling

marksto:
@oliyh Hi Oliver!

This one is a bit heavy, but it's an absolute minimum required to get the job done and to get it done well.

The problem is the multipart requests (support for which was recently added by me in #225) handling didn't play well with non-trivial data/types and request params coercion against the schema. For example, if your API schema has a numeric or array parameter, good luck sending those values ​​as the content of a multipart request (in the current implementation). You will either fail schema coercion if you prematurely convert the values ​​of these parameters to strings, or you will get an error about an unsupported content value if you try to pass them as is.

Moreover, the sets of supported content value types differ from one HTTP client to another. Finally, some clients fail to handle a multipart request with existing "Content-Type" header, either by throwing an error when sending the request (upon a `response-for` call, probably due to multiple conflicting header values) or [by not including a correct header with the `boundary` value set](https://github.com/dakrone/clj-http/pull/654) at all.

All this led me to a redesign that would defer decisions about converting multipart request content values ​​until the right moment (after the parameters have been successfully coerced), do it in a centralized and automated way and in accordance with HTTP clients' support of custom content types (while serving as a "common denominator" between all target clients), and eliminate the existing "Content-Type" request header, if any.

And the cool thing is, it works. And the uncompromising test coverage I had to add gives me the confidence to say so.

Hope this one will make it into the new version of the library!

Cheers,
Mark


core/src/martian/encoders.cljc
```diff
@@ -36,8 +37,15 @@
...
-     (mapv (fn [[k v]] {:name (name k) :content v}) body)))
+   (defn multipart-encode
+     ([body]
+      (mapv (fn [[k v]]
+              {:name (name k) :content (multipart/coerce-content v)})
+            body))
+     ([body pass-pred]
+      (mapv (fn [[k v]]
+              {:name (name k) :content (multipart/coerce-content v pass-pred)})
+            body))))
```
marksto, line 48: Here's an updated `multipart-encode` encoder fn for "multipart/form-data" requests, that now also does `:content` values coercion (so that they comply with the RFC specification). Allows the custom content types to "pass through" as is (the `pass-pred` parameter). Yet, it's backward compatible with the previous version.

core/src/martian/interceptors.cljc
```diff
@@ -83,22 +84,34 @@
...
+
+                                          multipart?
+                                          ;; NB: Luckily, all target HTTP clients — clj-http (but not lite), http-kit,
+                                          ;;     even hato and org.babashka/http-client — all support the same syntax.
+                                          (-> (set/rename-keys {:body :multipart})
+                                              (prepare-multipart-request-headers))
+
+                                          (and content-type (not multipart?))
+                                          (assoc-in [:headers "Content-Type"] content-type))]
+              (assoc ctx :request encoded-request)))})
```
marksto, line 114: This is one of the primary changes. Removes an existing "Content-Type" request header, if any, from _all_ multipart requests.

core/src/martian/multipart.cljc
```diff
@@ -0,0 +1,50 @@
+(ns martian.multipart
+  "An adaptor that prepares multipart content for all supported HTTP clients"
+  (:require [clojure.java.io :as io])
+  (:import (java.io File InputStream)))
+
+;; http-kit = {String | File, InputStream, byte[] | ByteBuffer! | ~Number!~}
+;; clj-http = {String | File, InputStream, byte[] | o.a.h.e.m.c.ContentBody}
+;; hato     = {String | File, InputStream, byte[] | URL, URI, Socket, Path!}
+;; bb/http  = {String | File, InputStream, byte[] | URL, URI, Socket, Path?}
+
```
marksto, line 10: A conceptual diagram of the content types supported by all target HTTP clients. Strings, common binary types, extra binary types (everything that can be coerced to `InputStream`) and custom types (only for `clj-http` and `http-kit`).

core/src/martian/multipart.cljc
```diff
@@ -0,0 +1,50 @@
...
+              (io/input-stream obj)
+              (catch Exception _ex
+                (str obj)))]
+
+           :clj
+           [(implements-make-input-stream? obj)
+            (io/input-stream obj)
+
+            :else
+            (str obj)])))))
```
marksto, line 50: This is the main logic for converting `:content` values into suitable types.

If we leave aside the bb specifics, it looks like this:

```clojure
(defn coerce-content
  ([obj]
   (coerce-content obj nil))
  ([obj pass-pred]
   (when (some? obj) ; let it fail downstream
     (cond
       (or (string? obj) ; <~ strings go as is
           (common-binary? obj) ; <~ as well as common binary types
           (and pass-pred (pass-pred obj))) ; <~ or custom types (if any)
       obj

       (implements-make-input-stream? obj) ; <~ coerce extra binary types
       (io/input-stream obj)

       :else
       (str obj))))) ; <~ otherwise, simply stringify the content value

```

The bb version is slower, but functionally equivalent.

P.S. Wrestling with Babashka is at times harder than it should be... 😩

core/src/martian/schema.cljc
```diff
@@ -135,26 +135,29 @@
...
+                           "uuid" (s/cond-pre s/Str s/Uuid)
                            s/Str)
     (= "integer" type)   s/Int
     (= "number" type)    s/Num
     (= "boolean" type)   s/Bool
     (= "date-time" type) s/Inst
-
-    :else
-    s/Any))
+    :else                s/Any))
```
marksto, line 160: Here's a missing "binary" format of the "string" type that is required for OpenAPI/Swagger specifications with "multipart/form-data" request parameters. CLJS schema is perfectly "String or `js/File`". For CLJ we have to postpone until the HTTP client-specific `coerce-content` fn call, so it's intentionally `s/Any` for the time of request params coercion.

core/test/martian/interceptors_test.cljc
```diff
@@ -71,9 +71,9 @@
                      (assoc (encoders/default-encoders)
                        "multipart/form-data" {:encode encoders/multipart-encode
                                               :as :multipart}))]
-             (is (= {:multipart [{:name "alpha" :content 12345}
+             (is (= {:multipart [{:name "alpha" :content "12345"}
                                  {:name "omega" :content "abc"}]
-                     :headers {"Content-Type" "multipart/form-data"}}
+                     :headers nil}
                     (:request ((:enter i) {:request {:body body}
                                            :handler {:consumes ["multipart/form-data"]}}))))))))))
```
marksto, line 78: Ooops, a "breaking" change. But no target HTTP client will be able to send this request anyway. `¯\_(ツ)_/¯`

marksto:
@oliyh I know that `+1,178 −125` may be too big of a change to dig through, so I did my best to highlight the key spots where important decisions were made. Hope this helps!

babashka-http-client/test/martian/babashka/http_client_test.clj
```diff
@@ -181,3 +197,168 @@
...
+                :url "http://localhost:8888/upload"
+                :multipart [{:name "string" :content "Howdy!"}]
+                :headers {"Accept" "application/json"}
+                :as :text}
+               (martian/request-for m :upload-data {:string "Howdy!"})))
+        (is (match?
+              {:version :http1.1
+               :status 200
+               :headers {:content-type "application/json;charset=utf-8"}
+               :body {:content-type multipart+boundary?
```
oliyh, line 217: We seem to have different behaviour here depending on whether version is `http-1.1` or `http1.1` which look semantically the same to me - would you mind explaining what the difference is here?

I also can't see how an identical function call is resulting in the two different `:version` values?

oliyh: Still confused by the inconsistency between `http-1.1` and `http1.1`

clj-http/test/martian/clj_http_test.clj
```diff
@@ -73,3 +93,224 @@
...
+                  :multipart [{:name "binary" :content tmp-file}]
+                  :headers {"Accept" "application/json"}
+                  :as :text}
+                 (martian/request-for m :upload-data {:binary tmp-file})))
+          (is (match?
+                {:status 200
+                 :headers {:content-type "application/json;charset=utf-8"}
+                 :body {:content-type multipart+boundary?
+                        :content-map {:binary (binary-content tmp-file)}}}
+                (martian/response-for m :upload-data {:binary tmp-file})))))
```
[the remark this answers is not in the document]
oliyh: Why should there not be a content-type in the martian-built request?

babashka-http-client/test/martian/babashka/http_client_test.clj
marksto: This is a `babashka-http-client`-specific thing (as you can see [here](https://github.com/oliyh/martian/blob/master/babashka-http-client/src/martian/babashka/http_client.clj#L28-L29)), so now I regret that I didn't remove it completely, but left strict equality here and there. Since this has nothing to do with the task at hand or the values ​​being tested, I will remove all `:version`s.

core/src/martian/interceptors.cljc
```diff
@@ -83,22 +84,34 @@
               (update ctx ::tc/queue #(into (into tc/queue i) %))
               ctx))})
 
+(defn- prepare-multipart-request-headers [request]
+  (update request :headers dissoc "Content-Type" "content-type" :content-type))
```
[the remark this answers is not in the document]
marksto: Fixed in https://github.com/oliyh/martian/pull/230/commits/c84a128d6b59574e3628d25bc304fc9d435b5d8e.

marksto:
@oliyh Please, have a look now. I reduced the new tests size by ~100 lines.

marksto:
@oliyh Please, have a look now. I reduced the new tests size by ~200 lines and extracted one more common helper fn, the `without-content-type?`, to have test checks more robust, specific and comprehensible. But I would still like to keep them isolated and preserve locality because of the number of client-specific discrepancies.

oliyh:
Great improvement in the tests, much more specific now and should be more robust too - thanks for taking the extra effort.

And once again thanks for your valuable contributions!

oliyh:
Yes, can do a release later today.

Cheers

Issue #234, opened 2025-07-03
HTTP client-specific output coercion issues


marksto:
@oliyh Hi Oliver!

This one is all about HTTP client-specific output (response body) coercion discrepancies (or, better put, variance), some of which cause subtle issues with/in Martian.

### Summary

| HTTP client | Request map key | Supported raw types | `Content-Type`-based auto-coercion (value) | Supported `Content-Type`s (matching values) | 
| ----------- | ----------- | ----------- | ----------- | ------------------------------------------------ |
| `hato` | `:as` | `nil`/`:default` (string)<br/>`:byte-array`<br/>`:stream` | yes (`:auto`) | `"application/json"` (`:json`, `:json-string-keys`, `:json-strict`, `:json-strict-string-keys`)<br/>`"application/edn"` (`:clojure`)<br/>`"application/transit+json"` (`:transit+json`)<br/>`"application/transit+msgpack"` (`:transit+msgpack`)<br/>`"text/*"` [only for charset decoding] | 
| `clj-http` | `:as` | `nil`/`:default` (string)<br/>`:byte-array`<br/>`:stream`<br/>`:reader`<br/>`"charset"` (string) | yes (`:auto`) | `"application/json"` (`:json`, `:json-string-keys`, `:json-strict`, `:json-strict-string-keys`)<br/>`"application/edn"`, `"application/clojure"` (`:clojure`)<br/>`"application/transit+json"` (`:transit+json`)<br/>`"application/transit+msgpack"` (`:transit+msgpack`)<br/>`"application/x-www-form-urlencoded"` (`:x-www-form-urlencoded`)<br/>`"text/xml"` (`:xml`) |
| `clj-http-lite` | `:as` | `nil` (default, string)<br/>`:byte-array`<br/>`:stream`<br/>`"charset"` (string)<br/>`:auto` (string) | no, `:auto` only affects the string charset | `"text/*"` [only for charset decoding] |
| `http-kit` | `:as` | [with automatic unzipping turned on]<br/>`:auto` (default)<br/>`:text` (string)<br/>`:byte-array`<br/>`:stream` | yes (`:auto`), but only raw types | — |
| `babashka-http-client` | `:as` | `nil`/`:string` (default)<br/>`:bytes` [sic!]<br/>`:stream` | no | — |
[...]

Issue #233, opened 2025-07-02
Babashka HTTP client has "application/transit+msgpack" in supported content types (BB/SCI)


marksto:
While it shouldn't be there, and Michael clearly notes it in the code, the "application/transit+msgpack" is still present, which can eventually lead to an error:

```
java.lang.NoClassDefFoundError: Could not initialize class org.msgpack.MessagePack
```

The `transit+msgpack` is not available when running in BB, but is available when running on the JVM.

Issue #189, opened 2023-10-04
json/decode is called twice with clj-http if api operation.response.content is "*/*"


felixdo:
Hi, thanks for this amazing tool. It's what I've been looking for a long time.

I'm consuming an api that has an operation with response.content set to "*/*", whereas it actually produces application/json. 
I'm using clj-http, and for this case, martian uses its { :as :auto } option. The server responds with content-type: application/json, so clj/http auto-coerces the result into a map, whereas the martian coercion interceptor still accepts a string, and then fails when it calls cheshire to decode. I wonder what the best way to mitigate this would be (apart from fixing the api spec).

felixdo:
Yes that would work, but not sure if it's the best solution. Maybe its not needed if we get it right earlier.

As a workaround, I just remove `:as :auto`  in another interceptor that I insert after interceptors/coerce-response. 

Maybe one could allow a customizable auto-encoder (e.g. via a special key in the encoders map)
I'm also not sure if there's really a valid use case to ever pass `:as :auto`, I would maybe consider removing it.
Then also.. you dont want to break existing clients, so maybe customizable auto-encoder is the way to go.

marksto:
Removing `:as :auto` makes total sense, since the Babashka HTTP client doesn't support this option. I had issues with it during the #230 development (https://github.com/oliyh/martian/pull/230/commits/ab49e72b91bc722e89dffda0ae5e4e748bc22834). I think this also relates to #58.

marksto:
Hi @felixdo! This should be fixed now. You should be able to test it after the upcoming major release.

Issue #58, opened 2018-10-04
Failure to load swagger.json if Content-Type is not application/json


polymeris:
My server is returning `swagger.json` with a `Content-Type: text/plain` header, which seems to confuse martian. It fails to register any endpoints, *silently*.

Not sure if this is something you want to fix or expected behavior, but I would suggest adding a warning, if possible.

polymeris:
Btw, I am not sure if `:as :json` here does anything. I think that's a `clj-http` option.
https://github.com/oliyh/martian/blob/master/cljs-http/src/martian/cljs_http.cljs#L29

oliyh:
Hi,

Thanks for reporting this. Yes, you're right - the `:as :json` doesn't seem to be an option in cljs-http, instead it's just looking for the `application/json` content type.

I think a warning could be ok but I would prefer it to be something that could be checked. I would like to have a spec for the Martian instance which could allow validation at creation time and then the developer could decide what to do.

I'll have a think about this, it's probably unlikely to happen soon though. PRs welcome :)

marksto:
> Yes, you're right - the `:as :json` doesn't seem to be an option in cljs-http, instead it's just looking for the `application/json` content type.

From what I can see now (years later), the `cljs-http` actually [do look for a similar property](https://github.com/r0man/cljs-http/blob/74b51788f6c27a4802a9d6aa89ba3512ac589fdf/src/cljs_http/core.cljs#L33-L43), namely the `:response-type` key.

This is a different issue though. Martian has been warning the user for a long time that there's no matching encoder for a given content type in the OpenAPI/Swagger definition.

But it would be really nice to have a built-in support for responses of `Content-Type: text/plain`. It's an easy thing to add/have.

marksto:
@oliyh Well, this one should still be open, as per my previous comment — the `"text/plain"` and `"application/octet-stream"` media types still do not get proper treatment with regards to OpenAPI JSON schema parsing. They should be made auto-available upon parsing, so there no warnings. Plus, the warning itself should be made tunable (as previously agreed in #214).

PR #235, opened 2025-07-03
Fix response coercion issues

### LINKED ISSUE #232: Response Content-Type header is not retrieved (properly)
@oliyh Hi Oliver!

The same discrepancy [we discussed earlier](https://github.com/oliyh/martian/pull/230#discussion_r2169411668) also applies to coercion.

| HTTP client | Content-Type header key in the response `:headers` map |
| ----------- | ------------------------------------------------------ |
| `clj-http` | `"Content-Type"` |
| `clj-http-lite` | `:content-type` |
| `http-kit` | `:content-type` |
| `hato` | `:content-type` |
| `babashka-http-client` | `:content-type` |
| `cljs-http` | `"content-type"` |

Sad but true — this very cause as well contributes to the open issue #189 and #58.

I have the fix for this issue (and some other response coercion-related ones) readily available locally.
But I am waiting for the #230 to be merged first.

Cheers,
Mark

marksto:
**In Scope:**

- #232
- #234
- #233
- #189
- #58
- #237
- #236

**Notes:**

These all are closely related, and it's super hard to fix them individually while preserving logical completeness, correctness and some basic compatibility. For instance, when one fixes #232, everything else starts to fall apart because an encoder for a specific "Content-Type" now gets detected and the "two negatives make an affirmative" rule stops working for us! 😫 

That said, I decided to ship everything together, but in a gradual way (which someone may call "TDD").

First, I introduce a missing test coverage for all target HTTP clients and supported content types (and the `"*/*"` OpenAPI/Swagger response content special case). I also put a number of ToDo items in tests for all the issues I've found. Please, note that the new tests for the issues in scope are expected to fail!

Second, implement and roll out all the fixes.

Third, augment the existing test coverage to nail down the results of the fix and/or to highlight potential compatibility issues. Here I don't add new test cases, but rather introduce additional checks for existing ones.

Finally, I clean up the tests code a bit after relaxing some constraints that are no longer needed; add test coverage for both ways of using `clj-http` and `hato` clients (with `:auto`-coercions on and off); and expose new Martian bootstrap options, namely `:use-client-output-coercion?` and `:async?`, for capable clients.

### DIFF
--- a/core/src/martian/encoders.cljc
+++ b/core/src/martian/encoders.cljc
@@ -1,6 +1,5 @@
 (ns martian.encoders
-  (:require [clojure.string :as string]
-            [clojure.walk :refer [keywordize-keys]]
+  (:require [clojure.string :as str]
             [cognitect.transit :as transit]
             [flatland.ordered.map :refer [ordered-map]]
             #?(:clj [cheshire.core :as json])
@@ -12,6 +11,8 @@
                 :clj [[ring.util.codec :as codec]]))
   #?(:clj (:import [java.io ByteArrayInputStream ByteArrayOutputStream InputStream PushbackReader])))
 
+;; NB: THIS NS ASSUMES THAT "UTF-8" IS THE CHARSET OF ALL ENCODED/DECODED VALUES!
+
 #?(:clj
    (defn as-bytes [obj]
      (cond
@@ -21,7 +22,8 @@
        (.getBytes ^String obj)
 
        (instance? InputStream obj)
-       (with-open [^InputStream is obj, baos (ByteArrayOutputStream.)]
+       (with-open [^InputStream is obj
+                   baos (ByteArrayOutputStream.)]
          (io/copy is baos)
          (.toByteArray baos))
 
@@ -75,11 +77,15 @@
   ([body]
    (json-decode body {:key-fn keyword}))
   ([body opts-or-fn]
+   ;; There's also `cheshire.parse/*use-bigdecimals?*` available
    (let [{:keys [key-fn array-coerce-fn]} (if (fn? opts-or-fn)
                                             {:key-fn opts-or-fn}
                                             opts-or-fn)]
-     #?(:clj  (json/parse-string body key-fn array-coerce-fn)
-        :cljs (when (and (string? body) (not (string/blank? body)))
+     #?(:clj  (if (string? body)
+                (json/parse-string body key-fn array-coerce-fn)
+                (with-open [rdr (io/reader body)]
+                  (json/parse-stream rdr key-fn array-coerce-fn)))
+        :cljs (when (and (string? body) (not (str/blank? body)))
                 (-> body
                     (js/JSON.parse)
                     (js->clj :keywordize-keys (boolean key-fn))))))))
@@ -118,7 +124,8 @@
 
 (defn form-decode [body]
   #?(:bb   nil
-     :clj  (keywordize-keys (codec/form-decode (as-string body)))
+     :clj  (let [params (codec/form-decode (as-string body))]
+             (update-keys (if (map? params) params {}) keyword))
      :cljs (let [params (js/URLSearchParams. body)]
              (reduce (fn [acc k]
                        (let [v (.getAll params k)]
--- a/core/src/martian/encoders.cljc
+++ b/core/src/martian/encoders.cljc
@@ -135,34 +135,47 @@
                      {}
                      (.keys params)))))
 
+(defn transit-encoder
+  [type transit-opts & kvs]
+  (conj {:encode #(transit-encode % type (:encode transit-opts))
+         :decode #(transit-decode % type (:decode transit-opts))}
+        (apply hash-map kvs)))
+
+(defn json-encoder
+  [json-opts & kvs]
+  (conj {:encode #(json-encode % (:encode json-opts))
+         :decode #(json-decode % (:decode json-opts))}
+        (apply hash-map kvs)))
+
+(defn edn-encoder
+  [edn-opts & kvs]
+  (conj {:encode #(edn-encode % (:encode edn-opts))
+         :decode #(edn-decode % (:decode edn-opts))}
+        (apply hash-map kvs)))
+
+(defn form-encoder
+  [& kvs]
+  (conj {:encode form-encode
+         :decode form-decode}
+        (apply hash-map kvs)))
+
 (defn default-encoders
   ([] (default-encoders keyword))
   ([opts-or-fn]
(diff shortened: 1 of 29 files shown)


marksto:
**The end of the third phase.**

All tests were successfully passed and meet actual expectations. All issues within the scope have been resolved. 👍

marksto:
@oliyh Hi Oliver!.. Sorry for another biggy one, but it was very hard for me to stop midway. 😅 

This one definitely requires a major version bump as a signal to end users. And it would be nice to have some release notes.

core/src/martian/encoders.cljc
```diff
@@ -1,40 +1,94 @@
 (ns martian.encoders
```
marksto, line 1: The primary change number 2 — the revamped `martian.encoders` namespace.

core/src/martian/interceptors.cljc
```diff
@@ -164,13 +229,16 @@
              ctx)}))
 
 (defn supported-content-types
```
marksto, line 231: The primary change number 4 — preserving encoders order in the `supported-content-types` fn.

marksto: I would rather keep the `:encodes` and `:decodes` values as (ordered) sets, to preserve the data type, but it turned out that `flatland.ordered.set` ns does not have CLJS support. Ah oh.

marksto:
During my self-review I found one subtle bug with a potential to backfire. The issue didn't happen for the default coercion settings (the `coerce-opts` of the `coerce-response` interceptor), but it could have happened for the user-modified ones. Fixed it and covered with extra tests.

marksto:
The final tuning comes down to reducing (essentially eliminating) any substantial impact on VCR-based tests for `clj-http` and `hato` clients. Double-checked everything locally. The only difference is `:as :string` (now) vs. `:as :text` (before) in the recorder `:request` (if any, for `clj-http`), which is fine — both values (legacy and current) will work the same in terms of response coercion. Most importantly, all recorded response bodies remain intact.

marksto:
Ready for review.

core/src/martian/interceptors.cljc
```diff
@@ -125,23 +128,31 @@
...
+      :leave (fn [{:keys [response coerce-as] :as ctx}]
+               ;; TODO: In some cases (`http-kit`) it may be necessary to decode an `:error :body`.
+               (let [content-type (when (:body response)
+                                    (get-content-type (:headers response)))]
+                 (if-not (or (encoding/skip-decoding? coerce-as content-type coerce-opts)
+                             (encoding/auto-coercion-by-client? coerce-as coerce-opts))
+                   (let [{:keys [decode]} (encoding/find-encoder encoders content-type)
+                         decoded-response (update response :body decode)]
+                     (assoc ctx :response decoded-response))
+                   ctx)))})))
```
marksto, line 155: The primary change number 1 — the `coerce-response` interceptor rectification.

core/src/martian/encoders.cljc
```diff
@@ -1,40 +1,94 @@
...
+#?(:clj
+   (defn as-string
+     ([obj]
+      (if (string? obj)
+        obj
+        (as-string obj "UTF-8")))
+     ([obj ^String charset]
+      (if (string? obj)
+        obj
+        (slurp obj :encoding charset)))))
```
marksto, line 48: The decoder fns are now more flexible thanks to `as-<raw-type>` coercion fns.
Their design was influenced by existing `clj-http`/`hato` utils implementation.

core/src/martian/encoders.cljc
```diff
@@ -47,46 +101,81 @@
...
+
+(defn edn-decode
+  [body opts]
+  (let [opts (merge {:eof nil} opts)]
+    #?(:clj  (if (string? body)
+               (edn/read-string opts body)
+               (with-open [rdr (io/reader body)
+                           pb-rdr (PushbackReader. rdr)]
+                 (edn/read opts pb-rdr)))
+       :cljs (edn/read-string opts body))))
```
marksto, line 118: New! Customisable `edn-encode`/`edn-decode` fns.

core/src/martian/encoding.cljc
```diff
@@ -1,25 +1,92 @@
 (ns martian.encoding
```
marksto, line 1: The primary change number 3 — the `martian.encoding` ns with new and simplified old logic.

core/src/martian/encoding.cljc
```diff
@@ -1,25 +1,92 @@
...
 (def auto-encoder
   {:encode identity
-   :decode identity
-   :as :auto})
+   :decode identity})
+
+(defn find-encoder [encoders media-type]
+  (or (when-some [type-subtype (get-type-subtype media-type)]
+        (get encoders type-subtype auto-encoder))
+      auto-encoder))
```
marksto, line 25: The old stuff is simplified and properly renamed. (As far as we use `martian.encoding` ns fns for both "Accept" and "Content-Type" encoders, they are about working with "media types" not "content types" per se.)

core/src/martian/interceptors.cljc
```diff
@@ -164,13 +175,16 @@
...
-          {:encodes #{}
-           :decodes #{}}
-          interceptors))
+  (-> (reduce (fn [acc interceptor]
+                (merge-with into acc (select-keys interceptor [:encodes :decodes])))
+              {:encodes []
+               :decodes []}
+              interceptors)
+      (update :encodes distinct)
+      (update :decodes distinct)))
```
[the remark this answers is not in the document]
marksto: I would rather keep the `:encodes` and `:decodes` values as (ordered) sets, to preserve the data type, but it turned out that `flatland.ordered.set` ns does not have CLJS support. Ah oh.

core/src/martian/encoding.cljc
marksto: The `auto-encoder` now goes without `:as :auto`, which makes it not so much an "auto" but rather a "missing" encoder.

oliyh:
Hi @marksto is this ready for review? I saw you kept pushing new stuff, didn't want to jump in too early.

Cheers

oliyh:
Great contribution, thank you once again @marksto.

I will do a release - would you be happy to help draft the release notes, and are there any changes you think should be made to the README (examples of new capabilities / options)?

Cheers

marksto:
@oliyh Hi Oliver! Sure thing, I can take some notes and, most importantly, collect all the potentially breaking changes into a list. Plus, I think, what's [In Scope](https://github.com/oliyh/martian/pull/235#issue-3198786981) could go as is into the release notes as well. I will update the README as well.

PR #238, opened 2025-07-16
Prepare for #235 release

### LINKED ISSUE #58: Failure to load swagger.json if Content-Type is not application/json

### LINKED ISSUE #189: json/decode is called twice with clj-http if api operation.response.content is "*/*"

### LINKED ISSUE #232: Response Content-Type header is not retrieved (properly)

### LINKED ISSUE #233: Babashka HTTP client has "application/transit+msgpack" in supported content types (BB/SCI)

### LINKED ISSUE #234: HTTP client-specific output coercion issues

### LINKED ISSUE #236: The `cljs-http-promise` module tests fail silently
Any failures, e.g. response coercion errors, are not visible in the test results. The false-positive successful passes are possible with all exceptions being silently swallowed by `promesa`.

### LINKED ISSUE #237: The `default-encoders` fn result must be ordered
Since there is logic (e.g. choosing a matching content type for the given encoders, determining suitable content type when parsing OpenAPI/Swagger definition, etc.) that depends on the resulting map being perfectly ordered — when in fact it is not, — we end up with undesired non-deterministic behaviour.

The `martian.interceptors/supported-content-types` fn has to also preserve the same (declaration) order in its `:encodes` and `:decodes` return values.

marksto:
@oliyh Hi Oliver!

I've prepared the changes from PR #235 for release.

The project README was updated, restructured, interlinked, and now comes with a ToC. Here I also managed to fix and/or update some code snippets with usage examples.

During this work, I also noticed another, previously uncovered use case of response coercion that (unsurprisingly) did not work with some real HTTP clients — for an encoder with the `:as :magic` entry. Also fixed. Supported by a new test case.

Feel free to browse through the commits one by one to review the latest changes. I tried to keep them atomic, as usual.

Proposed release notes:

> **WARNING: MAJOR BREAKING CHANGES**
>
> Every effort has been made to maintain backward compatibility, but the nature of the changes is such that the fixes introduced will inevitably change existing, incorrect behaviour.
>
> This release _breaks_:
> - Martian's built-in media type encoders (a map produced by the `martian.encoders/default-encoders` function) are now effectively ordered. This restores the expected behaviour, but may affect the client code if many (3 or more on JVM) custom encoders were added to the default ones, resulting in a change in the order of the map.
> - The `supported-content-types` function now preserves the encoders declaration order. This affects the OpenAPI definition parsing and is actually capable of breaking the code of some library clients. For instance, now, but only when applicable, the "application/transit+json" encoder may be chosen instead of the "application/json".
> - The `:as` value (used for Content-Type-based output auto-coercion by some target HTTP clients) of built-in media type encoders is now chosen correctly and takes into account the [specific capabilities of a particular HTTP client](https://github.com/oliyh/martian/issues/234). This will affect previously recorded VCR files for `clj-http` and `hato` clients, although the changes should be superficial, because the recorded request bodies and other details were not affected.
> - The legacy `martian.interceptors/default-encode-body` var was (intentionally) removed to signal to library customers that changes to their encoding/decoding code may be required.
>
> This release _fixes_:
> - Response Content-Type header is now properly retrieved for all supported HTTP clients, closes #232
> - All target HTTP clients now receive correct, client-specific config options for output (auto-)coercion, so all existing and potential response coercion issues should now go away, closes #234 and partially fixes #58
> - The `json/decode` no longer called twice in a row with `clj-http` in case the API `operation.response.content` is `"*/*"`, closes #189
> - The `martian.encoders/default-encoders` function result is now effectively ordered, closes #237
> - The "application/transit+msgpack" no longer sneaks into the `bb/http-client`'s supported content types on BB/SCI, closes #233
> - The `cljs-http-promise` module tests no longer fall silently, closes #236
> - [The `cljdoc` build error](https://cljdoc.org/builds/92006) for the Martian core module should now be fixed
>
> This release _improves_:
> - The project README structure, interlinking, copy (including brand new sections on default and custom behaviour) and usage examples (including the custom media type example fix)
>
> This release _adds_:
> - An ability to customise built-in Transit, JSON, and EDN encoders with a complete set of encoding/decoding options
> - Implicit raw types coercion functions in the `martian.encoders` namespace — `as-string`, `as-stream`, and `as-bytes` — so the encoder `:as` value works type-safely and in accordance with the design of HTTP client libraries.
> - Constructor functions for encoders of all built-in media types in the `martian.encoders` namespace — `transit-encoder`, `json-encoder`, `edn-encoder`, and `form-encoder` — which can be used directly to more easily customise the built-in encoders.
> - An ability to bootstrap Martian instance with custom request/response encoders and HTTP client-specific options, such as `:async?` and `:use-client-output-coercion?`

It would be cool to give this major, breaking release a descriptive name. And I would bump the major version, of course.

Cheers,
Mark


oliyh:
Hi,

I've done a `0.2.0` release (I also accidentally did a `0.1.35` release with the same content 🤕 )

Will set up the release on Github shortly

marksto:
@oliyh Cool, thank you! Should it also be on the [Releases](https://github.com/oliyh/martian/releases) page, shouldn't it? I see it's been quite a while since you last updated it.

oliyh:
Release created: https://github.com/oliyh/martian/releases/tag/0.2.0

Issue #216, opened 2025-03-03
`martian.cljs-http/bootstrap-openapi` sets incorrect `:api-root` value


enspritz:
Using `martian-cljs-http` and `martian-re-frame`. Discovered this upgrading from martian `0.1.16` to `0.1.30`.

```clojure
(martian.re-frame/init "/rest/openapi.json")
```
fetches and parses the OpenAPI document, storing the result in re-frame's `:db`. Upon inspecting the re-frame `:db`, the `:api-root` value appears incorrect:
```edn
{:db
 {:martian.re-frame/martian
  {:martian.re-frame/default-id
   {:m
    {:api-root ":///rest/" ...
```
All subsequent attempts to use the REST API through martian fail, due to HTTP requests sent to the wrong URL, such as:
```
/wherever/the/calling/document/resides.html/%3a///rest/some-call
```
If `%3a//` is interpreted as an encoded URL, that would be `://`, the same as in the erroneous `:api-root` value.

Skimming over the sources, `martian.openapi/base-url` looks of interest...

enspritz:
Workaround: Judicious use of `lambdaisland/uri` to combine the browser window location with the OpenAPI JSON document URL:
```clojure
(let [fully-qualified-url (lambdaisland.uri/join js/window.location.origin 
                                                 "/rest/openapi.json")]
  (martian.re-frame/init fully-qualified-url))
```

oliyh:
Hi,

Thanks for your solution, it looks sensible. The value you see in martian looks like it's failed to read the scheme, port and host from the browser. I'll try to find time to have a look, tempted to just use your solution.

enspritz:
The workaround might be a good defensive / backup / last-resort value. If only I had more time, I could dive into the logic of that function myself. Don't know the impact of modifying it..

Issue #168, opened 2023-03-15
Bootstrapping relative url for a swagger api from cljs doesn't work


oliyh:
e.g. `(bootstrap-swagger "/api/swagger.json")` ends up with an api root of `://` after it's been fed through `(openapi/base-url url server-url definition)`

marksto:
@oliyh Hi Oliver! As #253 addressed this issue, shouldn't we double-check that it now works as intended, cut off a new Martian release, and then close the ticket?

PR #253, opened 2026-03-05
Handle CLJS relative URLs for specs

bombaywalla:
When `bootstrap-openapi` is called with a relative URL in a CLJS environment, and the spec does not contain a `:servers` key, and no `:server-url` option is specified, then `bootstrap-openapi` should return a martian with an `:api-root` which is also a relative URL with the last path segment removed.

I believe this PR addresses #168 and #216.


bombaywalla:
Sorry, not following what you're saying.

I have similar (but not the same) code for testing cljs-http and cljs-http-promise.
Are you perhaps saying that I should only have tests for cljs-http and not for cljs-http-promise?
If so, I can go for that.

Issue #229, opened 2025-06-21
Feature request: constrain


eval-on-point:
Given this `example-openapi.yaml`:

```yaml
openapi: "3.0.1"
info:
components:
  parameters:
    offsetParam:  # <-- Arbitrary name for the definition that will be used to refer to it.
      # Not necessarily the same as the parameter name.
      in: query
      name: offset
      required: false
      schema:
        type: integer
        minimum: 0
      description: The number of items to skip before starting to collect the result set.
    limitParam:
      in: query
      name: limit
      required: false
      schema:
        type: integer
        minimum: 1
        maximum: 50
        default: 20
      description: The numbers of items to return.
paths:
  /users:
    get:
      operationId: getUsers
      summary: Gets a list of users.
      parameters:
        - $ref: '#/components/parameters/offsetParam'
        - $ref: '#/components/parameters/limitParam'
      responses:
        '200':
          description: OK
```
we do not get request validation on the bounds of the parameters:
```clojure
(let [client (martian-httpkit/bootstrap-openapi "droneup_client/example-openapi.yaml")]
  (martian/explore client :get-users)
  ;; => {:summary "Gets a list of users.",
  ;;     :parameters {{:k :offset} Int, {:k :limit} (default Int 20)},
  ;;     :returns {200 nil}}

  (martian/request-for client :get-users {:offset -1 :limit 51})
  ;; => {:method :get,
  ;;     :url "/users",
  ;;     :query-params {:offset -1, :limit 51},
  ;;     :as :auto}

  (martian/request-for client :get-users {:offset "string" :limit "string"})
  ;; => Execution error (ExceptionInfo) at schema-tools.coerce/coerce-or-error! (coerce.cljc:24).
  ;;    Could not coerce value to schema: {:offset (not (integer? string)), :limit (not (integer? string))}
  )
```

Can we add the integer constraints to the generated schema? I think this would be a breaking change if we turn it on by default, so I am asking if you would accept such a change before working on a PR.

Thanks you!

oliyh:
Hi,

Yes I think that would be a useful feature to have. One of the main use cases of martian is being able to validate offline that you are making the correct request, and this will help that.

Thanks!

eval-on-point:
[...]

PR #254, opened 2026-03-09
Added capability to specify min/max bounds in OAS.

bombaywalla:
For a schema with a type of "integer" or "number", you can now specify any or all of the following additional optional assertions: minimum, maximum, exclusiveMinimum, exclusiveMaximum.  As per OpenAPI 3.1.0.

With tests.

Addresses #229


bombaywalla:
I updated this PR to handle the recent changes in master, so there are no conflicts.
It should be ready to merge into master after review.

Issue #231, opened 2025-06-29
Advise for possible data loss caveat


felixdo-vorwerk:
Since some forms of OpenAPI schema are unsupported (allOf, anyOf), martian might produce only a partial internal schema representation from such API definitions. This can be catastrophic in combination with the "stc/map-filter-matcher" which just drops data. We should probably: Advise for this possibility on the caveats page and provide some instructions on how to drop the map-filter-matcher, it's not straightforward, at least not for me.

oliyh:
Hi @felixdo-vorwerk 

Thanks for pointing this out. I can add it to the caveats, and it should also be possible to control this with an option

felixdo-vorwerk:
Hey oliyh, nice, thanks.

For now, I just postprocess the martian, replacing all the body schemas with s/Any.

```
(defn replace-all-body-schemas [martian-def]
  (postwalk
    #(if (and (map? %) (:body-schema %))
       (assoc % :body-schema {:body s/Any})
       %)
    martian-def))
```

marksto:
> ... provide some instructions on how to drop the `map-filter-matcher`, it's not straightforward ...

TBH, it's a wrong thing to try to achieve in the first place. This `map-filter-matcher` is here for a reason — to shave off stuff that does not conform to a specific params schema. AFAICS, the problem is completely on the schema side, no? What gets produced for these "allOf" and "anyOf" definitions?

marksto:
Hi @felixdo! I think I didn't make myself clear enough, sorry for that.

The reason for using `map-filter-matcher` in the current implementation is simple. Martian accepts a single map of method `params`. The entries of this map then end up in different places — body, query, form, and headers. And since we don't want, say, body parameters to end up in the query string and/or request headers, the library need to somehow filter them out accordingly. That is why it coerces `params` against a set of closed schemas (one per params destination) and for each of them it discards non-conforming parameters (using the aforementioned `map-filter-matcher`).

Therefore, the data loss you experienced is due to an incomplete closed schema and also because the `params` are designed around a single map.

PR #258, opened 2026-03-27
Backend

oliyh:


marksto:
@oliyh Is this an Easter gift for us all? 😅 This already looks super cool! Godspeed!

core/src/martian/schema_backend.cljc
```diff
@@ -0,0 +1,50 @@
...
+
+  ;; Default value wrapping
+  (with-default-value [backend property s]
+    "Wraps schema s with a default value from property's :default key.
+     Returns s unchanged if :default is not present.")
+
+  ;; Collection format handling
+  (wrap-collection-format-schema [backend array-schema collection-format]
+    "Wraps an array schema to handle a specific collection format string
+     (e.g. 'csv', 'ssv'). Returns array-schema unchanged if not applicable.")
```
marksto, line 42: This seems to be the only place where impedance mismatch occurs. These "collection formats" are not inherent in schemas, right? Maybe we can decompose it into atomic methods instead — two predicates (`string-schema?`, `array-schema?`) and one constructor? Or, given that we've already allowed OpenAPI specifics to leak into this protocol (the `leaf-schema` method), is it fine to leave this method as is?

marksto:
Oliver, if you need any help porting the Plumatic Schema-specific part of the codebase I recently implemented in order to fully support the "parameter aliases" feature, just give me a nod. These are basically:

- the more or less self-contained [`martian.schema-tools`](https://github.com/oliyh/martian/blob/master/core/src/martian/schema_tools.cljc) ns, which is specific to Schema and is only used by
- the [`martian.parameter-aliases`](https://github.com/oliyh/martian/blob/master/core/src/martian/parameter_aliases.cljc) ns that exposes 3 functions: `alias-schema`, `unalias-data`, and `registry`

Hopefully, for Malli we won't have to re-implement `alias-schema`/`unalias-data` with custom pre-walks with path, but will rather use Malli schemas [built-in introspection capabilities](https://github.com/metosin/malli?tab=readme-ov-file#programming-with-schemas) (if they are sufficient enough, I didn't check yet, but it looks like we should be able to leverage `mu/subschemas`, `mu/path->in`, and friends). As for the `registry`, since Malli is fully supported by Babashka, we'll only need to implement a lazy variant (similar to the `LazyRegistry` instance).

core/src/martian/schema.cljc
```diff
@@ -100,30 +100,33 @@
...
     (sb/with-default-value backend param schema)))
 
+(defn- parameter-entries
+  "Turns a collection of swagger parameters into map schema entries."
+  [ref-lookup parameters backend]
+  (map (fn [{:keys [name required required?] :as param}]
+         {:key (keyword name)
+          :required? (boolean (or required?
+                                  (true? required)
+                                  (= "true" required)))
```
marksto, line 110: This `or`'s operands look a bit off.

core/src/martian/schema_backend.cljc
```diff
@@ -46,6 +46,20 @@
...
+  ;; Parameter aliases (idiomatic kebab-case keys)
+  (key-paths [backend schema]
+    "Returns a sequence of key paths (vectors of plain keys) covering the
+     schema itself (the [] path) and every map entry reachable within it.")
+  (aliases-at [backend schema idiomatic-path]
+    "Returns a map of idiomatic (kebab-case) keys to original keys for the
+     map level at idiomatic-path within schema, or nil when there are none.")
+  (alias-schema [backend aliases schema]
+    "Returns schema with the keys of it and its subschemas renamed to their
+     idiomatic (kebab-case) forms using the given aliases registry.")
```
marksto: "... forms using the given aliases registry" — there is no more `registry` here in the arg list.

core/src/martian/schema_backend.cljc
```diff
@@ -46,6 +46,20 @@
...
+  (merge-map-schemas [backend schemas]
+    "Merges several map schemas into a single map schema, with the entries
+     of later schemas winning.")
+
+  ;; Parameter aliases (idiomatic kebab-case keys)
+  (key-paths [backend schema]
+    "Returns a sequence of key paths (vectors of plain keys) covering the
+     schema itself (the [] path) and every map entry reachable within it.")
+  (aliases-at [backend schema idiomatic-path]
+    "Returns a map of idiomatic (kebab-case) keys to original keys for the
```
marksto: This "idiomatic" is more than kebab-case. Check out the `martian.schema-tools/->idiomatic` fn impl.

core/src/martian/spec.cljc
```diff
@@ -1,14 +1,15 @@
...
 
 (s/def ::api-root string?)
 
 (s/def ::method keyword?)
 (s/def ::route-name keyword?)
 (s/def ::path-parts (s/coll-of (s/or :literal string? :arg keyword?)))
 
-(s/def ::input-schema (s/nilable (s/and map? #(satisfies? schema/Schema %))))
+;; NB: The schema representation is owned by the schema backend (e.g. a map
+;;     for Plumatic Schema, a vector form for Malli), so no shape is imposed.
```
marksto: "a vector form for Malli" — it's [a bit trickier than that](https://github.com/metosin/malli#vector-syntax); e.g. `:string` (a kwd) is a Malli schema, or anything that was returned by a `malli.core/schema` fn call.

core/src/martian/backends/malli.cljc
```diff
@@ -0,0 +1,274 @@
...
+   Select it by passing `:schema-backend martian.backends.malli/backend` in
+   the opts of any bootstrap function."
+  (:require #?(:cljs [goog.Uri])
+            [clojure.string :as string]
+            [malli.core :as m]
+            [malli.error :as me]
+            [malli.transform :as mt]
+            [malli.util :as mu]
+            [martian.schema-backend :as sb]
+            [martian.schema-tools :as mst :refer [unalias-data]]))
```
marksto: How a Malli schema backend can use `martian.schema-tools`, which should be for (Plumatic) Schema?

core/src/martian/backends/malli.cljc
```diff
@@ -0,0 +1,274 @@
...
+  (int-schema [_] :int)
+
+  (map-schema [_ entries {:keys [open?]}]
+    (-> [:map]
+        (into (map (fn [{:keys [key required? schema]}]
+                     (if required?
+                       [key schema]
+                       [key {:optional true} schema])))
+              entries)
+        (cond-> open? (conj [::m/default :any]))))
```
marksto, line 146: There's no need in that line — `(cond-> open? (conj [::m/default :any]))` — maps are _open by default_ in Malli (and one needs to explicitly put `{:closed true}` property to have this different).

core/src/martian/backends/malli.cljc
```diff
@@ -0,0 +1,274 @@
...
+(defn- container-form? [form]
+  (and (vector? form) (contains? transparent-containers (first form))))
+
+(defn- map-form-entries
+  "Returns [key value-schema] pairs for the concrete entries of a :map form,
+   ignoring the ::m/default catch-all entry of open maps."
+  [form]
+  (for [entry (form-children form)
+        :when (and (vector? entry) (not= ::m/default (first entry)))]
+    [(first entry) (peek entry)]))
```
marksto: That looks brittle, tbh. Have you checked out [leveraging built-in Malli introspection capabilities](https://github.com/metosin/malli?tab=readme-ov-file#programming-with-schemas)? I mentioned them [above](https://github.com/oliyh/martian/pull/258#issuecomment-4145301305) some time ago.

marksto (commented):
Hi Oliver! It's nice to see you are really pushing hard on that pluggable schema backend feature. I reviewed the current set of changes. A great job overall! Please, find my comments below. I hope they'll help to improve what's already there.

core/src/martian/backends/malli.cljc
oliyh: Thanks for all this @marksto - I've been letting Claude loose on this and was planning to review it at the end. This was Fable 5 - not so impressive!

core/src/martian/backends/malli.cljc
```diff
@@ -201,6 +133,10 @@
   (int-schema [_] :int)
 
   (map-schema [_ entries {:keys [open?]}]
+    ;; Malli maps are open for validation already, but the default transformer
+    ;; strips undeclared keys on coercion. An open map (OpenAPI's
+    ;; `additionalProperties`) adds a ::m/default catch-all entry so those keys
+    ;; survive coercion instead of being stripped.
```
marksto, line 139: Well, that's a good point overall, and I understand it, but — have you read [my reply to this issue](https://github.com/oliyh/martian/issues/231#issuecomment-3270248897)? It feels to me that it's against the current Martian design. Maybe it's a good time to revisit it, idk. I'm just flagging this for your careful consideration...

core/src/martian/schema.cljc
```diff
@@ -105,7 +105,7 @@
   (map (fn [{:keys [name required required?] :as param}]
          {:key (keyword name)
           :required? (boolean (or required?
-                                  (true? required)
+                                  (= true required)
                                   (= "true" required)))
```
marksto, line 110: Here I still don't get it, even after the "fix" — why do even have both `required` and `required?` params? How they differ and where do they come from? Can they be normalized into a single param?

core/src/martian/schema_backend.cljc
marksto, line 42: And I still believe we are mixing things up here. Yes, we now have a seemingly full set of protocol methods for schema providers, but why are we still infusing this protocol with OpenAPI specifics? That was my point [before](https://github.com/oliyh/martian/pull/258/changes/c689abf265adc78e07ff303cd317f1d3735264d4#r3003093457), and it remains.

marksto: I'm not saying that it's a wrong thing to do — maybe the design is sound, I don't know — but it puzzles me.

marksto (commented):
Re-reviewed. That's way much better now — better matches Malli's capabilities and also reserves all existing parameter aliases-specific details, both in code and in docs/comment. A few remaining questions I still have, if you don't mind, are:

core/src/martian/backends/malli.cljc
marksto: @oliyh No worries, mate, it's my pleasure to assist here and, hopefully, be of some real help, not just random nagging.

Btw, I find this Fable model to be really good in such deep code research/refactorings. Not perfect, sure, but indeed superior to everything I myself have tried before. It's great to see it's helping us move this mountain! Malli support is way overdue...

core/src/martian/backends/malli.cljc
oliyh: Not directly relevant to this change, as it's an existing open question.
I'm inclined to say if the body is an open map then extra params are allowed to propagate to the body. If query params is an open map then extra params are allowed to propagate to the query. If they are both open... not sure!

Without this, martian becomes unusable, or at least requires some ugly escape hatches, if you want to use the extra params explicitly allowed by the schema.

core/src/martian/schema.cljc
oliyh: I would like to normalise these. I think it stems from "required" being a boolean or a string "true" in OpenAPI. Pre-existing though before this backend change

core/src/martian/schema_backend.cljc
oliyh: They are not "inherent" in the schema, as in it doesn't directly influence the schema. But the method of attaching the metadata for later use is implementation specific. I'm not quite sure how else to do it, unless we have a "with-meta" protocol. How would your idea look?

I'm not too concerned about this overall - the point was to decouple martian from the schema it uses. This is achieved. If we've gone a bit overboard and pushed some stuff too far down it makes it less pure but it doesn't matter functionally.

core/src/martian/schema.cljc
```diff
@@ -20,10 +21,10 @@
...
   plumatic/default-coercion-matcher)
 
-(defn get-backend
-  "Returns the schema backend from opts, defaulting to the Plumatic backend."
-  [opts]
-  (get opts :schema-backend plumatic/backend))
+(def get-backend
+  "Returns the schema backend from opts, defaulting to the Plumatic backend.
+   Re-exported from `martian.backends` for backward compatibility."
+  backends/get-backend)
```
marksto, line 27: Is it still needed? What for? The code section says "Backward-compatible re-exports from the Plumatic backend", but this new API was never published.

core/test/martian/backends/malli_test.cljc
```diff
@@ -0,0 +1,277 @@
...
+      (is (= [:map
+              [:foo-bar :string]
+              [:foo {:optional true} [:maybe [:map [:doo-dar {:optional true} [:maybe :string]]]]]
+              [:tags [:sequential [:map [:tag-name :string]]]]]
+             (aliases/alias-schema backend (aliases/registry backend s) s))))
+
+    (testing "unalias-data renames data keys back to the schema's keys"
+      (is (= {:fooBar "a" :foo {:dooDar "b"} :tags [{:tagName "c"}]}
+             (aliases/unalias-data (aliases/registry backend s)
+                                   {:foo-bar "a" :foo {:doo-dar "b"} :tags [{:tag-name "c"}]}))))))
```
marksto, line 277: This is a ridiculously small test coverage compared to `martian.parameter-aliases-test` that I managed to compile back in the day. I wish it could serve as a basis for Malli schema backend, almost ad verbum.

core/src/martian/backends/malli.cljc
marksto: > Not directly relevant to this change, as it's an existing open question.

Yeah, let's postpone this redesign. We definitely can have pluggable schema backends first, then decide on how to approach that one properly.

> ;; Malli maps are open for validation already, but the default transformer
> ;; strips undeclared keys on coercion.

The more I think about it, the less I believe that we do actually need to copy that behavior of Plumatic Schema at all. Even Malli [had problems](https://github.com/metosin/malli/issues/930) with this `mt/strip-extra-keys-transformer` at some point.

It's fair to say that `::m/default` catch-all entry is not smth that is absolutely necessary. The pitfall can be avoided by marking all non-closed map schemas as explicitly open (`{:closed false}`) or simply by avoiding the very use of `mt/strip-extra-keys-transformer` in the `default-transformer` stack.

core/src/martian/schema.cljc
marksto: Yeah, I get it, but I'm mostly bothered with `required?` (with question mark) vs. `required` (w/o question mark). They shouldn't both be needed at the same time, should they?

core/src/martian/schema_backend.cljc
marksto: Fair. Still, my internal purist is not super happy about this impedance mismatch that can be addressed.

> How would your idea look?

1. Have a schema backend-specific String schema similar to how `(int-schema [_] ...)` is declared.
2. Have a schema backend-specific ways to add/get "metadata" (in a broad sense) to/from a schema.

The `seq-schema` is already there. Composition with `item-schema` should work. Adding/getting metadata works the same as it is done in the current impl. With Plumatic Schema it's literal `meta`/`vary-meta`. With Malli it goes through type properties (`{...}`/`m/properties`).

The algorithm itself is generic — I'm talking `format-coll-coercion-matcher`/`seq->string`/`format->separator` vs. `collection-format-decoder`/`collection-format->separator` — they just hook up a bit differently, but the formatting logic is the same and can be represented by a shared, generalized fn.

Am I missing something here?

core/src/martian/core.cljc
```diff
@@ -207,8 +221,19 @@
...
    - `:coercion-matcher`   — a unary fn of schema used for parameters coercion;
-                             defaults to the `default-coercion-matcher`;
+                             defaults to the `default-coercion-matcher`. Applies
+                             to the Plumatic backend only — under the Malli
+                             backend use the `:transformer` option instead;
+   - `:transformer`        — a Malli transformer used for parameters coercion;
+                             defaults to the Malli backend's `default-transformer`.
+                             Applies to the Malli backend only — under the
+                             Plumatic backend use the `:coercion-matcher` option
+                             instead;
```
marksto, line 236: I'd put all such options under `:schema-backend-opts` here and in 2 other places.

